PR #3, opened 2017-03-15
Add README.md

steveklabnik:


README.md
```diff
@@ -0,0 +1,70 @@
...
+To build the Nomicon:
+
+```bash
+$ mdbook build
+```
+
+The output will be in the `book` subdirectory. To check it out, open it in
+your web browser.
+
+_Firefox:_
```
frewsxcv: It'd be cool if `mdbook build` had a `--open` flag like `rustdoc` so we didn't have to spell out how to open a world wide web browser from a command line

steveklabnik: iirc it does?

steveklabnik: i copied this from the book, so, it might be wrong just because of that

Issue #2, opened 2017-03-15
Expand lifetimes section


vext01:
Hi,

I'd like to expand the lifetimes section of the rustonomicon with the example (and Lukas's answer) to the question in [this stack overflow post](http://stackoverflow.com/questions/42807679/clarify-the-meaning-of-binding-two-references-to-differently-scoped-referents-to/42808207#42808207).

If you'd welcome this change I will do the legwork and raise the PR, but I want to check that such a change is welcomed before investing the time.

(I'm a Rust beginner, and a first time contributor to rust docs).

Thanks

steveklabnik:
Hi @vext01 !

Yes, I'd love some PRs for the rustinomicon, including this stuff. Just send one in, that'd be great.

I'm going to give this a close since the question is answered 😄

vext01:
Thanks Steve.

I was going to suggest that my PR could be marked to fix this bug, but as you wish :)

vext01:
@steveklabnik Do you guys have a formal name for the process of deciding the shortest scope to satisfy referents from different scopes (as described in the SO post)? I'm wondering what I should call the new section...

Also, how does one build the nomicon? Is it an mdbook project?

steveklabnik:
> I was going to suggest that my PR could be marked to fix this bug, but as you wish :)

Let's do that!

> Do you guys have a formal name for the process of deciding the shortest scope to satisfy references to referents from different scopes (as described in the SO post)?

I am not 100% sure, /cc @nikomatsakis

> Also, how does one build the nomicon? Is it an mdbook project?

yes. I really need to add a README....

PR #4, opened 2017-03-16
Overhaul Lifetimes Section (was "Help Needed: Extend lifetimes section")

vext01:
Hi,

As mentioned in #2, I'm trying to better document the lifetimes system of Rust, as it's the part all beginners trip on.

I was trying to combine the collective knowledge of what the nomicon already tells us about desugraing to see implicit lifetimes, and the intuition given by the answer of my stack overflow question (http://stackoverflow.com/questions/42807679/clarify-the-meaning-of-binding-two-references-to-differently-scoped-referents-to/42808207#42808207).

All was going well until I tried to show a counter example where the lifetime constraints imposed by a function signature can't be satisfied (the last rust code block added in this PR). At this point something has gone wrong, since I am unable to come to the same conclusion as the rust compiler. The compiler rejects the program, the intuitive method I am trying to explain within this PR does not reject the program!

Can an expert take a read of the stack overflow question, then the change in this PR and comment on where I have gone wrong?

 * Am I de-sugaring correctly?
 * Are the rules for converting lifetimes correct?
 * Are the rules for detecting an inconsistency correct?

I really wonder if there is a better way to express how the compiler is assigning and checking lifetimes, as it's not very intuitive as it stands. Before reading the nomicon, I've always though of the lifetime of a reference as the lifetime of the referent, but the nomicon seems to explain it in another way.

Cheers


vext01:
(excuse the typos in the commit messages, I will squash this all into one nice commit at the end)

steveklabnik:
> (excuse the typos in the commit messages, I will squash this all into one nice commit at the end)

No worries at all!

vext01:
I think the problem here is the rules for "converting" lifetimes. I don't think it helps to think in that way.

Given the de-sugared code:
```
fn main() {
    'a {
        let s1 = String::from("short");
        'b {
            let res: &'b str;
            'c {
                let s2 = String::from("a long long long string");
                'd {
                    // Annonymous scope for the borrows of s1 and s2
                    // Assigning to the outer scope causes s1 and s2 to have 'b
                    res: &'b = shortest(&'d s1, &'d s2);
                    println!("{}", res);
                }
            }
        }
    }
}

fn shortest<'k>(x: &'k str, y: &'k str) -> &'k str {
    ...
```

The way my colleague see this is, the lifetimes are inconsistent because:

 * The sig of `shortest()` tells us the return value has the same lifetime as the two args.
 * The args were marked `&'d` by de-sugaring, so we must show that the lifetime we annotated on res, i.e. `&'d`, is consistent with the lifetime we annotated earlier: `&'b`.
 * To be consistent, `&'d` would have to live longer than `&'d`, which it does not, so after the call res may refer to a freed value.

Therefore the program is inconsistent. Makes sense to me, but there might be a more intuitive way of thinking about this.

vext01:
Alright. Can someone with jedi rust skills check the most recent version.

I've borrowed (no pun intended) the style of [this post](http://stackoverflow.com/questions/42637911/how-can-this-instance-seemingly-outlive-its-own-parameter-lifetime/42639814#42639814), which makes a great deal of sense to me.

If you guys like this, then I would then need to go back and revise the prior sections to mark references with the lifetimes of their referents, and I would probably want to adjust the flow a bit too.

There's also the issue of (I think) move semantics giving rise to additional gotchas (since they limit the lifetimes of values), but I think I'd like to tackle that as a separate PR.

@steveklabnik Are you able to review this, or do you know someone who could?

Thanks.

vext01:
(I'm unable to test under mdbook, due to https://github.com/ogham/rust-ansi-term/issues/27, so I've been just pushing and viewing in GH ;) )

steveklabnik:
@vext01 I will try to soon, but in addition, @aturon and I have been talking about having lang team / docs team coordination to work on the reference; we may want that for the nomicon as well.

vext01:
OK great. Let me know if I can be of any help, as I'm keen to learn about this stuff.

vext01:
Sorry to persist, but has anyone at moz had a chance to review this?

steveklabnik:
I have not, sorry :(

aturon:
cc @Gankro

src/lifetimes.md
```diff
@@ -213,3 +213,204 @@ totally ok*, because it keeps us from spending all day explaining our program
 to the compiler. However it does mean that several programs that are totally
 correct with respect to Rust's *true* semantics are rejected because lifetimes
 are too dumb.
+
+# Inter-procedural Borrow Checking of Function Arguments
+
+Consider the following program:
+
+```rust,ignore
```
frewsxcv: Is there any reason why this is `ignore`? Seems like it should be compilable and runable, no?

frewsxcv: Same with some of the other examples here

src/lifetimes.md
```diff
@@ -213,3 +213,204 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+*caller* (`main`) are consistent with the lifetimes in the signature of the
+*callee* (`print_shortest`).
+
+The signature of `print_shortest` simply requires that both of it's arguments
+have the same lifetime (because both arguments are marked with the same
+lifetime identifier in the signature). If in `main` we had done:
+
+```rust,ignore
+print_shortest(&s1, &s1);
+
```
frewsxcv: Nit: unnecessary newline here

src/lifetimes.md
```diff
@@ -213,3 +213,204 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+                let s2 = String::from("a long long long string");
+                res: &'res: str = shortest(&'s1 s1, &'s2 s2);
+                println!("{}", res);
+            }
+        }
+    }
+}
+```
+
+Again, at the call-site of `shortest` the comipiler needs to check the
```
frewsxcv: Typo: comipiler

src/lifetimes.md
```diff
@@ -213,3 +213,204 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+consistency of the arguments in the caller with the signature of the callee.
+The signature of `shortest` fisrt says that the two reference arguments have
```
frewsxcv: Typo: fisrt

src/lifetimes.md
```diff
@@ -213,3 +213,204 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+fn shortest<'k>(x: &'k str, y: &'k str) {
+    if x.len() < y.len() {
+        println!("{}", x);
+    } else {
+        println!("{}", y);
+    }
+}
+```
+
+`print_shortest` simply prints the shorter of its two pass-by-reference
```
steveklabnik: "simply" should be taken out

src/lifetimes.md
```diff
@@ -213,3 +213,204 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+
+> A function argument of type `&'p T` can be coerced with an argument of type
+> `&'q T` if the lifetime of `&'p T` is equal or longer than `&'q T`.
+
+At our call site, the type of the arguments are `&'s1 str` and `&'s2 str`, and
+we know that  a `&'s1 str'` outlives an `&'s2 str`, so we can substitute `&'s1
+s1` with `&'s2 s2`. After this both arguments are of lifetime `&'s2` and the
+call-site is consistent with the signature of `print_shortest`.
+
+[More formally, the basis for the above rule is in *type variance*. Under this
```
steveklabnik: this would be all in `>`s, not `[]`.

src/lifetimes.md
```diff
@@ -213,3 +213,204 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+model, you would consider a longer lifetime a sub-type of a shorter lifetime,
+and for function arguments to be *co-variant*. However, an understanding of
+variance isn't strictly required to understand the Rust borrow checker. We've
+tried here to instead to explain using intuitive terminlolgy.]
+
+# Inter-procedural Borrow Checking of Function Return Values
```
steveklabnik: this should be a smaller heading than #

vext01:
Thanks everyone for the comments.

Formatting aside, is what I have written correct? Notice how, when de-sugaring, I initialise references with the lifetime of the referent, whereas the previous examples do something slightly different, often using a narrower lifetime.

Of course, a narrower lifetime can be fine (and probably beneficial, if it can be shown valid), but I wonder if that can be considered a separate "optimisation" that we don't mention here. I think it's important that we have a single intuitive narrative for explaining this to new users. I can't understate this. Lifetimes are what make Rust unique, but are also very hard to grasp (I'm still grasping myself to be honest).

So assuming my de-sugaring is correct (I hope someone checked, because I'm really very new to this), should I go back and revise the earlier examples to fit this style too?

Thanks

vext01:
Anyone? I was hoping this kind of PR would be prioritised, since the borrow checker is the one thing that no newcomer understands for a long time.

Mark-Simulacrum:
Sorry about that! We'll try to get you a review soon.

cc @steveklabnik @frewsxcv

steveklabnik:
Yes, sorry; I've been traveling _and_ have not yet quite worked out how
lang + docs should collaborate on the reference; dealing with all of this
is one of my goals for the upcoming week.

steveklabnik:
Hey @vext01;

so, i was talking with @nikomatsakis and/or @aturon about how we can work through things like this; I think the plan will be to eventually go through everything with them, rather than looking at individual PRs. So, I don't think we need actual lang team signoff here.

I *think* that this is all correct, but it's not my area of specialty. It's unclear to me who should actually sign off on it for that reason :/ Maybe @Gankro himself?

src/lifetimes.md
```diff
@@ -213,3 +213,204 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+```
+
+Now we see that the references passed as arguments to `print_shortest`
+actually have different lifetimes (and thus a different type!) since the values
+they refer to were introduced in different scopes. At the call site of
+`print_shortest` the compiler must now check that the lifetimes in the
+*caller* (`main`) are consistent with the lifetimes in the signature of the
+*callee* (`print_shortest`).
+
+The signature of `print_shortest` simply requires that both of it's arguments
```
Gankro: Kill all "simply"s

Gankro:
So I'm a bit concerned that there seems to be some redundancy with other sections. For instance https://doc.rust-lang.org/nomicon/subtyping.html discusses this merging behaviour a bit already (though it comes later). 

I need to re-read this chapter to see what gets discussed where.

vext01:
I've fixed the comments on my new prose.

Outstanding topics we need to discuss (quoting my earlier comments):

> Notice how, when de-sugaring, I initialise references with the lifetime of the referent, whereas the previous examples do something slightly different, often using a narrower lifetime.

> There's also the issue of (I think) move semantics giving rise to additional gotchas (since they limit the lifetimes of values), but I think I'd like to tackle that as a separate PR.

We might decide to do these in separate PRs, as this PR will start to bitrot.

I'm glad the rust team are starting to think about how to deal with docs. I can't emphasize enough the importance of documenting rust, since it has radical new ideas which even lifelong programmers (including myself) are struggling to put to work. You need to make it as easy as possible to learn rust, and documentation is all you've got! :)

If you keep me in the loop, I am happy to review/comment of PRs. I hope I can offer the "newbie's-eye-view".

vext01:
Fixed build.

steveklabnik:
@Gankro when you're happy with this, I think I'm happy with this 😄

src/lifetimes.md
```diff
@@ -213,3 +213,205 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+> if the lifetime of `&'r T` is equal or shorter than `&'s T`.
+
+To make our program compile, we would have to subsitute `res: &'res` for `res:
+&'s2`, but we can't since `&'res` in fact out-lives `&'s2`. This program is
+inconsistent and the compiler rightfully rejects the program because we
+try make a reference (`res`) which outlives one of the values it may refer to
+(`s2`).
+
+> Formally, function return values are said to be *contravariant*, the opposite
+> of *covariant*.
```
Rufflewind: I feel the argument/return position might be a bit of a red herring here.

I wouldn’t use the term contravariant to refer to return types, at least not in this sense.  They appear so because [a return type was compared to an argument type](https://github.com/rust-lang-nursery/nomicon/pull/4/files?diff=unified#diff-d12e10c55aa9b4bcd76634d6be48ab5eR362), but if you compared a return type to another return type they would appear covariant just as well!

---

The lifetime coercion (subtyping) rule is simply `&'a T -> &'b T` provided that `'a` outlives `'b`.  Argument/return type doesn’t play a role here.\*  The compiler is allowed to *shrink* the lifetime of a reference, but never extend it.  Formally speaking, reference types are said to be “lifetime-covariant”.

Another rule is that an object must always outlive its binding.  This means you can’t, in a variable of a larger scope, store a reference to an object of a smaller scope:

```rust
'r_scope {
    let r: &'r_ref;
    'inner {
        let inner = String::new();
        r = &'r_ref inner; // doesn't work because 'r_ref ≥ 'r_scope > 'inner
                           // yet 'r_ref ≤ 'inner
    }
}
```

From the first example, you learned that `'k` has to be `'s2` (or shorter).  This remains true for the second example.

In the second example, the returned reference has lifetime `'k`.  When it gets stored in `res`, the compiler may shrink `'k` if necessary.  Therefore the lifetime `'res_ref` of the stored reference must be `'k` or shorter.  Here, I used `'res_ref` to distinguish between the lifetime of the reference `res: &'res_ref str`, and the lifetime of the scope itself `'res_scope { let res; ... }`.

From the structure of the code itself, we know that the scope of `s2` is contained fully within that of `res`.  Because an object must always outlive its binding, it follows that `'res_ref` must outlive `'res_scope`, which is strictly longer than `'s2`.

If we combine these three facts together,

  - `'k` must be `'s2` or shorter
  - `'res_ref` must be `'k` or shorter
  - `'res_ref` must be strictly longer than `'s2`

we see that there can be no solution that satisfies all three constraints :(

There is a way to fix it though, by making `'s2` and `'res_scope` equal:

```rust
fn main() {
    let s1 = String::from("short");
    let (res, s2);
    s2 = String::from("a long long long string");
    res = shortest(&s1, &s2);
    println!("{}", res);
}
```

--

\* Due to variance, the rule changes for more complicated for composite types like `Cell<&'a ()>` or `fn(&'a ()) -> &'b ()` but that doesn't apply here.

ltratt: > I wouldn’t use the term contravariant to refer to return types, at least not in this sense. They appear so because a return type was compared to an argument type, but if you compared a return type to another return type they would appear covariant just as well!

I'm not sure I understand what you mean. Can you expand on this?

Rufflewind: Compare these two statements:

Rule I:

> A function **argument** of type `&'p T` can be coerced with an 
**argument** of type `&'q T` if the lifetime of `&'p T` is equal or longer than `&'q T`.

Rule II:

> The **return value** of a function `&'r T` can be converted to an **argument** `&'s T` if the lifetime of `&'r T` is equal or shorter than `&'s T`.

Rule I compares **arguments** vs **argument**.  Rule II is **return value** vs **argument**.  If you compare things in opposite positions, you will always the opposite result, so Rule II is just a corollary to Rule I.

Moreover, I don't think "co-/contravariant" is the appropriate term for this.   Formally, we say `F` is covariant in its parameter if and only if `A ⊆ B` implies `F<A> ⊆ F<B>`, and contravariant if and only if `A ⊆ B` implies `F<A> ⊇ F<B>`.  There is no mention of argument vs return types at all.

ltratt: I agree that the language "The return value of a function `&'r T` can be converted to an argument" feels a bit odd.

However, I personally think that the type co/contra-variance is a useful thing to point out to people a) because it's surprisingly unintuitive (indeed, there are real programming languages that have got this wrong in their type systems) b) because that can be used as a hook to explain the relationship to lifetimes. However, one doesn't necessarily need to use those words provided one gets the concept across somehow. What I think the document is trying to say is "if a function takes a parameter of type T and returns a type U, you can pass a (non-strict) supertype T' and receive back a (non-strict) supertype U' " and then making the link between that and the length of lifetimes.

Rufflewind: Did you mean "if a function takes T and returns U, you can pass a *subtype* T' and supertype U'"?

ltratt: Quite correct -- mea culpa!

Rufflewind: Here's my concrete suggestion for this PR:

1. Rewrite Rule I (L289-290) to say something along the lines of "A reference can always be shrunk to one of a shorter lifetime.  In other words, `&'a T` can be implicitly converted to `&'b T` as long as `'a` outlives `'b`."

2. Replace "for function arguments to be co-variant" (L299) with "a reference `&'a T` is always covariant in its lifetime parameter `'a`."

3. Replace Rule II and adjacent paragraphs (L348-372) with something like:

    > Again, at the call-site of `shortest` the compiler needs to check the consistency of the arguments in the caller with the signature of the callee.  The signature of `shortest` says that all three references must have the same lifetime `'k`, so we have to find a lifetime `'k` such that:
    >
    >   - `&s1: &'s1 str` can be converted to the first argument `&'k str`,
    >   - `&s2: &'s2 str` can be converted to the second argument `&'k str`, and
    >   - the return value `&'k str` can be converted to `res: &'res str`
    >
    > ```rust,ignore
    > res: &'res = shortest::<'k>(&'s1 s1, &'s2 s2);
    >           //                ^^^^^^^
    >           //               's1 -> 'k ^^^^^^^
    >           //                        's2 -> 'k
    >           // ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    >           //           'k -> 'res
    > ```
    >
    > This leads to three requirements:
    >
    >   - `'s1` must outlive `'k`, 
    >   - `'s2` must outlive `'k`, and
    >   - `'k` must outlive `'res`.
    >
    > However,  `'res` *strictly outlives* `'s2` by virtue of being a larger scope that encloses `s2`, so it's impossible to satisfy these conditions.  The borrow checker rightfully rejects our program because we are making a reference (`res`) which outlives one of the values it may refer to (`s2`)."

vext01: OK. I'll make these edits (and see if it's understandable to a beginner) when i find some time.

Thanks for your comments Rufflewind.

steveklabnik: @vext01 any chance you can find the time sometime soon?

vext01: I'll try to find time this week.

I've learnt more Rust since I last looked at this, so I may have some edits of my own.

vext01:
I've implemented Rufflewind's recommended changes, and read the whole thing over.

I shortened a fair amount of my prose, as I felt I had been rambling a bit.

Let me know what you think.

vext01:
btw, I killed the stuff on variance, as I felt it was confusing and vague.

Rufflewind:
I think the revised version look great!

vext01:
Shall I squash?

vext01:
ping

vext01:
I think someone should make it their new year's resolution to merge this :)

src/lifetimes.md
```diff
@@ -213,3 +213,142 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+Then consistency is trivially proven, since both arguments would have
+the same lifetime `&'s1` at the call-site. However, for our example, the
+arguments have different lifetimes. We don't want Rust to reject the program
+because it actually is safe. Instead the compiler uses some rules for
+converting between lifetimes whilst retaining referential safety. The first
+such rule is as follows:
+
+> A reference can always be shrunk to one of a shorter lifetime. In other
+> words, `&'a T` can be implicitly converted to `&'b T` as long as `'a`
+outlives `'b.`
```
QuietMisdreavus: Should this line have a `>` on it too? (Github's rendering adds it to the blockquote just fine, but i'd rather have it all neat. `>_>`)

src/lifetimes.md
```diff
@@ -213,3 +213,142 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+        's2 {
+            let s2 = String::from("a long long long string");
+            print_shortest(&'s1 s1, &'s2 s2);
+        }
+    }
+}
+```
+
+(for brevity, we don't show the third implicit scope that would be introduced
+to limit to lifetimes of the borrows in the call to `print_shortest`)
```
Gankro: "the" lifetime.

Regardless, It seems like this can either be cut, or is Actually Very Important?

src/lifetimes.md
```diff
@@ -213,3 +213,142 @@ totally ok*, because it keeps us from spending all day explaining our program
 to the compiler. However it does mean that several programs that are totally
 correct with respect to Rust's *true* semantics are rejected because lifetimes
 are too dumb.
+
+# Inter-procedural Borrow Checking
```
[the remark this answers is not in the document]
Gankro: This is also maybe a Bad Title, because borrowchecking is specifically *not* interprocedural.

src/lifetimes.md
```diff
@@ -213,3 +213,142 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+
+(for brevity, we don't show the third implicit scope that would be introduced
+to limit to lifetimes of the borrows in the call to `print_shortest`)
+
+Now we see that the references passed as arguments to `print_shortest`
+actually have different lifetimes (and thus a different type!) since the values
+they refer to were introduced in different scopes. At the call site of
+`print_shortest` the compiler must now check that the lifetimes in the
+*caller* (`main`) are consistent with the lifetimes in the signature of the
+*callee* (`print_shortest`).
```
Gankro: This is not how the lifetime system works. It's a constraint solver; they never have "wrong" lifetimes, and the lifetimes aren't checked if they're "right". It just tries to solve the system of constraints and either reports success or failure.

vext01:
Hi @Gankro!

Thanks for looking at my PR.

> I just don't understand the motivation for this change; or at least it doesn't feel like the nomicon is the right place for this.

I started working on this because I didn't understand how lifetimes work behind the scenes. Although I've gotten better at Rust since then, there are still times where I "brute force" may way past cryptic lifetime errors offered up by the compiler. And it's not just me, I occasionally see my friends and colleagues (experienced programmers by the way) struggling in the same way.

I think this is indicative of a problem with the mental model which non-expert Rust programmers have when writing Rust code. I also think that we need to find an intuitive way to describe what is really going on here, because guessing how to fix these weird errors just doesn't satisfy my curiosity.

> It's extending the scoped lifetime illustration which has never been great.

Yes agreed.

> All discussions of lifetimes are a mess as a result of non-lexical lifetimes, and this entire section probably needs to be rewritten to focus on the new system (and this addition just exacerbates the situation, I think)

Yeah, I've seen the RFCs about non-lexical lifetimes.

> I apologize greatly for having kicked this ostensibly simple review down the road for so long.
> ...
>  It absolutely sucks that you have been strung along this long by my negligence, but it happened. I'm sorry.
> ...
> As such I'm inclined to cut our losses and just close this.

Don't sweat it! I completely understand. Sometimes these things don't work out. And in this instance I'd rather we get it right than cling on pretending to my precious code is helping :)

Before I close this, can I recommend that we make an issue on this (or another) repo that basically says "find a more intuitive way to explain lifetimes to users". If you let me know which repo, I will happily raise the issue.

You might consider describing the constraint solver in this repo though? I am happy to help reviewing if you can write the prose. [I have experience with linear constraint solving and SAT].

Thanks :smile:

src/lifetimes.md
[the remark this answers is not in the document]
vext01: Perhaps it's the wrong term, but I was referring to the way that the arguments at call-sites must be checked against a function definition, so in some way lifetimes do cross the function boundary. If you see what I mean?

vext01:
Ping?

vext01:
I'm going to close this as there seems to be a lack of interest/time.

PR #8, opened 2017-04-04
Document `#[repr(u8)] enum` to enable LLVM optimizations

geofft:
See rust-lang/rust#27303 and rust-lang/libc#48 for discussion of the
optimization, or just try compiling `unsafe {free(malloc(16))}` with various
Rust definitions of `void *` in the FFI binding.

### DIFF
--- a/src/other-reprs.md
+++ b/src/other-reprs.md
@@ -37,6 +37,13 @@ enum representation in C is implementation defined, so this is really a "best
 guess". In particular, this may be incorrect when the C code of interest is
 compiled with certain flags.
 
+* "C-like" enums with `repr(C)` or `repr(u*)` still may not be set to an
+integer value without a corresponding variant, even though this is
+permitted behavior in C or C++. It is undefined behavior to (unsafely)
+construct an instance of an enum that does not match one of its
+variants. (This allows exhaustive matches to continue to be written and
+compiled as normal.)
+
 
 
 # repr(u8), repr(u16), repr(u32), repr(u64)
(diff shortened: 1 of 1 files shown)


Gankro:
This is a bit unclear to me as written. I would maybe zoom out from LLVM a bit and focus on the ABI implications of C-like enums: it will be treated as an integer, and not a struct-containing-an-integer.

CC @mystor

geofft:
Oh, hm, is a `#[repr(u8)] enum Foo {Bar(())}` represented as a struct, and thus passed on the stack in ABIs where structs are unconditionally on the stack? Then yeah, that's worth documenting too! Let me test that and then rework this paragraph, thanks.

mystor:
It might also be good to note that controlling the `#[repr]` of the enum doesn't change the set of legal values, and that unlike C/C++ enums, it is undefined behaviour to have a `#[repr(C)] enum Foo { A = 0, B = 1 }` with the value `10`.

geofft:
OK, I've rewritten it to take the above comments into account.

This has drifted a fair bit from my original subject; the specific optimization LLVM does with `malloc` and `free` (and particular, that LLVM considers `void *` to be a `* u8`) is more than I can fit in a passing comment. However, the comments on `std::os::raw::c_void` and `libc::c_void`, _plus_ the paragraph here about the enum needing to be C-like, are probably clear enough.

ubsan:
> If a C-like enum is missing a `repr(u*)` or `repr(C)`, it is treated like a one-member struct for ABI compatibility purposes.

This seems like a silly guarantee to make, since they should *not* be treated as a one-member struct for ABI purposes.

steveklabnik:
I have no idea what's right here...

mystor:
I generally agree with @ubsan That any properties of `repr(rust)` enums, with the potential exception of `Option<&T>` and `Option<Box<T>>`, should not be documented. Any internal implementation details of them should remain internal.

Gankro:
Agreed, the only property of repr(rust) that's guaranteed is that `OptionLike<&T>` has the same abi as `&T` and `*const T`.

steveklabnik:
Great. @geofft care to make these changes?

geofft:
Oh yeah, good points about not documenting the exact properties of `repr(rust)`. Thanks!

I've updated the pull request. This doesn't seem like the right area to discuss `OptionLike`, so I just removed all discussion of `repr(rust)` representation, but I can add a bit about that if that's helpful.

steveklabnik:
Let's :shipit: . Thanks!

PR #21, opened 2017-05-18
Add FFI chapter from the first edition of TRPL to nomicon

projektir:
The FFI page has been copied almost verbatim from the first edition of TRPL in the master branch of [rust-lang/book](https://github.com/rust-lang/book).

I've made changes to two links that would otherwise be broken due to TRPL being a directory lower than nomicon: see 2nd commit. Unfortunately, the links will be broken in the nomicon if it's built by itself outside of [rust-lang/rust](https://github.com/rust-lang/rust).


steveklabnik:
Thank you!

> Unfortunately, the links will be broken in the nomicon if it's built by itself outside of rust-lang/rust.

Yeah, there's nothing much to do about this, sadly.

PR #26, opened 2017-05-27
subtyping: clarify the concreteness of 'a in example

Rufflewind:
This matters because if `'a` were free to choose, then the argument would be trivial.  But in reality the argument also holds if `'a` was already chosen, which is what makes lifetimes special.


steveklabnik:
Thanks!

PR #27, opened 2017-05-30
rewrite references.md

Gankro:
This replaces the mostly pointless hand-waiving model with an honest declaration of "no one knows" and a discussion of why aliasing matters.

### DIFF
--- a/src/references.md
+++ b/src/references.md
@@ -1,12 +1,5 @@
 # References
 
-This section gives a high-level view of the memory model that *all* Rust
-programs must satisfy to be correct. Safe code is statically verified
-to obey this model by the borrow checker. Unsafe code may go above
-and beyond the borrow checker while still satisfying this model. The borrow
-checker may also be extended to allow more programs to compile, as long as
-this more fundamental model is satisfied.
-
 There are two kinds of reference:
 
 * Shared reference: `&`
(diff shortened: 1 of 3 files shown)


Gankro:
cc @ubsan

Gankro:
[Rendered](https://github.com/Gankro/nomicon/blob/2f44731b0a8ffa28758f5ebf1e1600ae456b5cc0/src/references.md)

aturon:
cc @ralfjung

src/references.md
```diff
@@ -17,161 +17,140 @@ Which obey the following rules:
 * A reference cannot outlive its referent
 * A mutable reference cannot be aliased
 
-That's it. That's the whole model. Of course, we should probably define
-what *aliased* means. To define aliasing, we must define the notion of
-*paths* and *liveness*.
+That's it. That's the whole model references follow.
 
+Of course, we should probably define what *aliased* means. An aliased--
```
mgattozzi: This seems oddly cut off and I had to read it a few times to get what the implied meaning was. It might just be best to say Rust doesn't have an aliasing model but let's talk about what aliasing is.

src/references.md
```diff
@@ -17,161 +17,140 @@ Which obey the following rules:
...
+    if *input > 5 {
+        *output *= 2;
+    }
+}
+```
 
+We would *like* to be able to optimize it to the following function:
 
+```rust
+fn compute(input: &u32, output: &mut u32) {
```
mgattozzi: I was a bit confused as to how this worked but it [does](https://is.gd/6MovUg)

src/references.md
```diff
@@ -17,161 +17,140 @@ Which obey the following rules:
...
+    let cached_input = *input; // keep input in a register (probably)
```
mgattozzi: Maybe remove the probably? I think it's implied here that this is an optimization we want made and this isn't a LIR or ASM representation so you can't show the register being kept.

src/references.md
```diff
@@ -17,161 +17,140 @@ Which obey the following rules:
...
+Specifically, we must concern ourselves with the existence of this degenerate
+input: `compute(&x, &mut x)`. With this input, we would get this execution:
 
-# Liveness
+```rust
+                    // *input == *output == 20 (just some large value)
+if *input > 10 {    // true  (input == 20)
+    *output = 1;
+}
+if *input > 5 {     // false (input == 1)
```
mgattozzi: If *input is 20 how is it now 1?

src/references.md
```diff
@@ -17,161 +17,140 @@ Which obey the following rules:
...
-heap. Box, Vec, String, and HashMap, are examples of types which uniquely
-own data on the heap.
+# Aliasing
 
-Unfortunately, data on the heap is not *always* uniquely owned. Rc for instance
-introduces a notion of *shared* ownership. Shared ownership of a value means
-there is no unique path to it. A value with no unique path limits what we can do
-with it.
+Variables and pointers are generally said to *alias* if they refer to the same
+location in memory.
```
RalfJung: Just a nit:
It doesn't have to be the same location. Pointer to (unaligned) integers also alias if they overlap. I wouldn't usually say this matters, but the next sentence says this is supposed to be the "broadest definition possible", so well ;)

src/references.md
```diff
@@ -17,161 +17,140 @@ Which obey the following rules:
...
+`input > 10`.
 
-It is also possible to reborrow into *multiple* mutable references, as long as
-they are *disjoint*: no reference is an ancestor of another. Rust
-explicitly enables this to be done with disjoint struct fields, because
-disjointness can be statically proven:
+The key thing to know about aliasing is that writes are the only thing that
+really matter. For instance, we have little concern for aliasing in this
+modified version of our function, because we've moved the writes to a temporary
+variable:
```
RalfJung: That doesn't sound entirely right to me. Consider the following:
```rust
fn foo(x: &i32) -> i32 {
  let y = *x;
  some_unknown_function();
  return y;
}
```
It should be possible to re-order the load and the function call, without knowing anything about that function (other than it being safe).
Now, one may argue that ultimately, this is about whether the unknown code writes to `x` or not. Still, saying that reads "don't really matter" seems to be an overstatement to me.

Gankro: The idea I was trying to get across is that reads only matter *insofar as* they relate to writes. If there are no writes, we don't care if there are reads (which is the concern with some_unknown_function).

src/references.md
Gankro: Ah in my mind a pointer points to a region of memory, so two unaligned pointers have overlapping regions, and a slice overlaps with any pointer-to-an-element.

RalfJung: I can see that. However, in that case, there is no "single location" in memory that they point to, which the wording in the text somewhat implies -- "refer to *the* same location" makes it sound like "the location a pointer points to" is a well-defined concept, whereas really it is a set of locations.

Gankro:
I've pushed a new commit that addresses the comments, I think.

TODO: adjust index to link the new page in

Gankro:
* [references.md rendered](https://github.com/Gankro/nomicon/blob/reref/src/references.md)
* [new page (that follows it) rendered](https://github.com/Gankro/nomicon/blob/reref/src/aliasing.md)

RalfJung:
Thanks, looks good to me!

Gankro:
I think this is ready to land?

steveklabnik:
Sounds good to me!

PR #30, opened 2017-07-01
Update compiler output in §3.9

pablocouto:
This updates the compiler output in §3.9, based on `rustc 1.18.0 (03fc9d622 2017-06-06)`¹.

¹[Rust Playground](https://is.gd/uGoSfd).


steveklabnik:
Thank you!

PR #32, opened 2017-07-14
Update for https://github.com/rust-lang/rust/pull/42959

SimonSapin:

PR #33, opened 2017-07-15
Update for https://github.com/rust-lang/rust/pull/42959

SimonSapin:


SimonSapin:
(This is the same as #32, but into a temporary branch so that the commit is reachable from the submodule in rust-lang/rust, so that Travis-CI can test https://github.com/rust-lang/rust/pull/42959.)

PR #40, opened 2017-10-18
Expand first-in-book use of certain abbreviations

hegza:
I recently read the nomicon to delve deeper into the language. I noticed that a couple of abbreviations were never explained. These unexplained, although common abbreviations can trip a reader, as at least ABI was not immediately clear to me.

### DIFF
--- a/src/other-reprs.md
+++ b/src/other-reprs.md
@@ -37,10 +37,10 @@ still consumes a byte of space.
   difference from a struct is that the fields aren’t named.
 
 * This is equivalent to one of `repr(u*)` (see the next section) for enums. The
-chosen size is the default enum size for the target platform's C ABI. Note that
-enum representation in C is implementation defined, so this is really a "best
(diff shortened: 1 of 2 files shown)


Gankro:
These are interesting cases, because at least in my mind, these aren't even really acronyms anymore. Rather they have become words on their own, in the same class as LLVM, GNU, CPU, GPU, etc.

Gonna check what some others think, but I'm inclined to merge this regardless. 

(Although I almost want to flip them to be like "FFI (Foreign Function Interface)", since the acronym is really the more important thing!)

hegza:
Done.

Part of why I worry about this is that when some people get here from higher level languages, these kinds of tutorials are the first thing they run into while learning about low-level things. These kinds of things increase the cognitive load. Of course, the expanded acronyms might be irritating or weird for someone more experienced, but that's probably why it's the first instance only that is expanded.

Gankro:
Thanks!

PR #43, opened 2017-11-20
Update for field reordering

bennyyip:
https://github.com/rust-lang/rust/blob/master/RELEASES.md
https://github.com/rust-lang/rust/pull/40377

### DIFF
--- a/src/repr-rust.md
+++ b/src/repr-rust.md
@@ -110,11 +110,6 @@ struct Foo<u32, u16> {
 The latter case quite simply wastes space. An optimal use of space therefore
 requires different monomorphizations to have *different field orderings*.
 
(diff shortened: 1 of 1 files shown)


Gankro:
Hey sorry for the delay! I think it would be better to just remove the note altogether? I think I just added this to avoid getting complaints about "why doesn't the compiler do this the book said so!"

what do you think?

bennyyip:
Great idea! Moreover, maybe we should replace it with something like: "Rust compiler does field reordering when not using repr(c)", make it more clear to the reader.

Gankro:
I think that's the whole point of this section, (and the one on repr(C)), so I don't think that will be necessary.

bennyyip:
Ok, I removed the notes.

Issue #29, opened 2017-06-24
FFI section suggests dangerous practice of using empty opaque type


RalfJung:
The FFI section of the Nomicon currently suggests to use an empty type for "opaque pointer types". That seems pretty dangerous. This came up in the discussion at <https://github.com/rust-lang/rfcs/pull/1861>, and also in <https://internals.rust-lang.org/t/recent-change-to-make-exhaustiveness-and-uninhabited-types-play-nicer-together/4602> many were not happy with using the empty type here.

Until we have proper opaque types, the safer suggestion seems to be do use a ZST with a private field, rather than an empty type. Then at least the type is not actively lying.

sfackler:
How is this dangerous?

RalfJung:
with `x: &Empty`, `match *x {}` could well be (no final decision has been bade yet) insta-UB if ever executed.

Actually, just passing a `&Empty` to another function is not really different from passing aliasing `&mut u32` to functions like `mem::swap` -- the promise made by the type is violated. It is not unlikely for the unsafe code guidelines process to decide that doing either of these is UB.

sfackler:
`&Empty` is not okay, but you're working with `*mut Empty` and `*const Empty` in FFI.

RalfJung:
> `&Empty` is not okay,

Glad we agree on that. :)

> you're working with *mut Empty and *const Empty in FFI.

While these types are fine, I am worried that people will quickly move to variables or function arguments of type `&Empty` in the wrappers they build around their FFI code.

This happened e.g. in <https://github.com/briansmith/ring/>, where `BIGNUM` [used to be](https://github.com/briansmith/ring/commit/6a030e31caf444fdcc2fcb2738140821c09118ed) an empty enum, and you can indeed [find `&BIGNUM`](https://github.com/briansmith/ring/blob/a67775a252a220578d642be175ee837ed1e3702e/src/rsa/bigint.rs#L593) in the code.

SimonSapin:
I’ve opened https://github.com/rust-lang-nursery/nomicon/pull/44 to recommend zero-size structs with a private field, instead.

https://github.com/rust-lang/rust/pull/45225#issuecomment-346493804 is another case of Bad Things happening with empty enums used in types that are not "impossible".

PR #44, opened 2017-11-23
Don’t recommend empty enums for opaque types

SimonSapin:
Fix https://github.com/rust-lang-nursery/nomicon/issues/29

### DIFF
--- a/src/ffi.md
+++ b/src/ffi.md
@@ -737,11 +737,11 @@ void foo(struct Foo *arg);
 void bar(struct Bar *arg);
 ```
 
-To do this in Rust, let’s create our own opaque types with `enum`:
+To do this in Rust, let’s create our own opaque types:
 
 ```rust
-pub enum Foo {}
-pub enum Bar {}
+#[repr(C)] pub struct Foo { private: [u8; 0] }
+#[repr(C)] pub struct Bar { private: [u8; 0] }
 
 extern "C" {
     pub fn foo(arg: *mut Foo);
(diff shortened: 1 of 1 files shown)


SimonSapin:
CC @skade

Gankro:
I literally didn't know this chapter was added??

SimonSapin:
@Gankro It seems to be imported from the Book’s first edition: https://doc.rust-lang.org/book/first-edition/ffi.html

src/ffi.md
```diff
@@ -737,11 +737,11 @@ void foo(struct Foo *arg);
...
 ```
 
-To do this in Rust, let’s create our own opaque types with `enum`:
+To do this in Rust, let’s create our own opaque types:
 
 ```rust
-pub enum Foo {}
-pub enum Bar {}
+#[repr(C)] pub struct Foo { private: [u8; 0] }
+#[repr(C)] pub struct Bar { private: [u8; 0] }
```
Gankro: I think this should be `_priv:` to avoid unused field warnings, and to match stdlib style.

I also don't really understand the point of repr(C) here? Silencing the lint?

SimonSapin: Yes, rustc warns without it. Same with `[u8; 0]` over `()`.

Gankro:
(heading to work) we should explain what each of the 3 weird things in this type accomplish, if we don't already.

RalfJung:
Given that empty types have been recommended here previously, and may still be recommended in various blogs posts or so out there, maybe there should be an explicit section explaining that empty enums should NOT be used?

Gankro:
ping @SimonSapin what do we want to do about this

RalfJung:
I opened https://github.com/rust-lang-nursery/nomicon/pull/82 to address the remaining concerns.

SimonSapin:
Oops, sorry I let this slip through the cracks. Thanks Ralf for taking over. #82 looks good so let’s close for that one.

PR #35, opened 2017-08-02
Update drop-related stuff based on improvements to the language

Gankro:
* ManuallyDrop now exists 🎉
* needs_drop is a thing (soon)
* drop check has changed *considerably*
  * there is now a section dedicated to the eyepatch (warning against its use)
* most references to drop check have been eliminated, as it's no longer an opt-out concern
* Unique replaced with Shared and `*mut`, as Unique is unlikely to be stabilized and unnecessary for safety
* misc drive-by cleanups

PR #47, opened 2017-12-11
Fix RFC 1327 hyperlink in dropck

dlrobertson:
- Fix hyperlink to RFC 1327 in the section on the dropck.
 - Fix other nits

### DIFF
--- a/src/dropck.md
+++ b/src/dropck.md
@@ -94,7 +94,7 @@ error: `days` does not live long enough
 error: aborting due to previous error
 ```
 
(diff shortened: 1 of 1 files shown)


dlrobertson:
This may not be relevant depending on when #35 will be merged.

Gankro:
Isn't the link exactly the same?

dlrobertson:
Yeah, but it was improperly formatted. See https://doc.rust-lang.org/nomicon/dropck.html#an-escape-hatch

PR #48, opened 2017-12-22
The `unique` feature is renamed `ptr_internals`

SimonSapin:
This PR and https://github.com/rust-lang/rust/pull/46952 mutually depend on each other.

### DIFF
--- a/src/casts.md
+++ b/src/casts.md
@@ -28,7 +28,7 @@ primitive:
  * `*T as integer`
  * `integer as *T`
  * `number as number`
- * `C-like-enum as integer`
+ * `field-less enum as integer`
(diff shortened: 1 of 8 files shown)


SimonSapin:
@Gankro you said that `Unique` is destined to never being stabilized and only staying as an implementation detail of the standard library. This rename is on that basis.

Perhaps sections of the Nomicon using `Unique` should eventually be rewritten to use `NonNull`? I suppose this depends on whether you want to explain how `std` works, or how one would go about writing their own collection.

SimonSapin:
With today’s Nightly this should now pass tests if you hit the Retry button on Travis-CI.

PR #60, opened 2018-03-13
Clarify non-nullable pointer optimization in repr(C) section

### LINKED ISSUE #59: Clarify FFI-safety of Option<extern "C" fn>
Relatively new to Rust, and reading through the Nomicon to make sure I build a new FFI module safely. I noticed one omission in https://github.com/rust-lang-nursery/nomicon/blob/master/src/other-reprs.md.

#13 added language to call out that `Option<&T>` is FFI-safe as long as `*const T` is FFI-safe. Based on https://github.com/rust-lang/rust/issues/40913, it seems that the preferred way of expressing nullable function pointers (returned via FFI) is using `Option<fn>` (or I guess `Option<extern "C" fn>` since function pointers set by C code are presumably also C-ABI).

If my understanding is correct, shall we add this to the Nomicon? I'm happy to send a PR if that's desired.

Also, are `Option<&T>` and `Option<extern "C" fn>` an exhaustive list of `repr(C)` FFI-safe enums with fields, or are there other exceptions to document as well?

ramosbugs:
Resolves #59

### DIFF
--- a/src/other-reprs.md
+++ b/src/other-reprs.md
@@ -26,12 +26,11 @@ still consumes a byte of space.
 * DST pointers (fat pointers), tuples, and enums with fields are not a concept
   in C, and as such are never FFI-safe.
 
-* As an exception to the rule on enum with data, `Option<&T>` is
-  FFI-safe if `*const T` is FFI-safe, and they have the same
-  representation, using the null-pointer optimization: `Some<&T>` is
-  represented as the pointer, and `None` is represented as a null
-  pointer.  (This rule applies to any enum defined the same way as
-  libcore's `Option` type, and using the default repr.)
+* If `T` is an [FFI-safe non-nullable pointer
+  type](ffi.html#the-nullable-pointer-optimization),
+  `Option<T>` is guaranteed to have the same layout and ABI as `T` and is
+  therefore also FFI-safe. As of this writing, this covers `&`, `&mut`,
+  `Box`, and function pointers, all of which can never be null.
 
 * Tuple structs are like structs with regards to `repr(C)`, as the only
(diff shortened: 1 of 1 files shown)


Gankro:
Looks good, just need to double-check that Box is Actually FFI-Safe, as this is quite surprising to hear

ramosbugs:
Thanks for the review!

The [Nullable Pointer Optimization](https://doc.rust-lang.org/nightly/nomicon/ffi.html#the-nullable-pointer-optimization) subsection under FFI calls out boxes since they can't hold null pointers. This makes it seem like they should be FFI-safe since the representation in memory would be the same as for a raw pointer (iiuc). I'm far from an authority on this, so please correct me if I'm wrong!

One potential distinction I see between `Box<T>` and the other listed types is that since a `Box` holds Rust-owned data (which is freed using Rust's allocator), it's not suitable to _receive_ pointers from non-Rust code. However, it should be able to _provide_ pointers from Rust to non-Rust code, right?

Gankro:
There are subtle annoying details about ABI here. Specifically, pointers are a specific kind of thing in ABIs, and a struct containing a pointer *does not* have the same one. I expect we currently do match the ABI, but I'm uncertain that we guarantee it. (why would we?)

Gankro:
Discussing with @eddyb, we agree we could guarantee the ABI of Box but don't have a strong motivation to do so. All we guarantee about `Box` is that `Option<Box>` has the same *layout* as `Box`.

ramosbugs:
Makes sense. I can't see any motivation for guaranteeing that ABI either. Updated the PR.

PR #61, opened 2018-04-10
Use the new GlobalAlloc trait

SimonSapin:
This is part of https://github.com/rust-lang/rust/pull/49669.

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -28,7 +28,7 @@ For instance, a custom implementation of `Box` might write `Drop` like this:
 ```rust
 #![feature(ptr_internals, allocator_api, unique)]
 
-use std::heap::{Heap, Alloc, Layout};
+use std::alloc::{Global, GlobalAlloc, Layout, Void};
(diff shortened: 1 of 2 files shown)


src/destructors.md
```diff
@@ -38,7 +38,7 @@ impl<T> Drop for Box<T> {
     fn drop(&mut self) {
         unsafe {
             drop_in_place(self.ptr.as_ptr());
-            Heap.dealloc(self.ptr.as_ptr() as *mut u8, Layout::new::<T>())
+            Global.dealloc(self.ptr.as_ptr() as *mut Void, Layout::new::<T>())
```
Gankro: ugh whose idea was it to make Void a thing in native rust code

SimonSapin: Oh sorry, I should have waited before opening this since we’re probably gonna rename Void before landing. Not sure to what, though.

SimonSapin:
Closing for now while we figure out what to do about `Void`.

PR #67, opened 2018-05-02
Admit that contravariance exists and clean up the subtyping section

Gankro:


Gankro:
cc @Havvy @nikomatsakis

Havvy:
I read through it and didn't see anything glaringly wrong, but this is also the first time I'm reading about covariance/contravariance.

So instead of directed feedback, here's a retrospective: I read through it twice.

The first time was as the intended audience, in which I learned the material being taught. The concepts flowed naturally and I learned what variance actually meant. I don't like the usage of "covariant" and "contravariant" (I'd rather subtypeable/supertypeable/uhhh...), as it feels too academic, but it's what every language uses, so 🤷‍♂️ . That conflicts of variance means invariant came naturally after seeing so many examples of types being co/contravariant and why. So overall, this is 🎊 🎊 🎊 . That said, I glossed over the example of variance combination as it just looked so busy.

The second time through, I looked for spelling mistakes and other copy-editing things. The only thing I really found is that the bullet points explaining co/contravariance combinations should have periods at the end because they are full sentences. And even then, it's not really necessary because they are bullets.

Gankro:
signed off on the theory by ajeffery

Issue #70, opened 2018-06-12
Implementing Vec final code does not compile


izderadicka:
The final example does not compile (just clicking the run button), with following messages:
```
   Compiling playground v0.0.1 (file:///playground)
warning: use of deprecated item 'std::heap::Heap': type renamed to `Global`
 --> src/main.rs:9:32
  |
9 | use std::heap::{Alloc, Layout, Heap};
  |                                ^^^^
  |
  = note: #[warn(deprecated)] on by default

warning: use of deprecated item 'std::heap::Heap': type renamed to `Global`
  --> src/main.rs:34:27
   |
34 |                 let ptr = Heap.alloc(Layout::array::<T>(1).unwrap());
   |                           ^^^^

warning: use of deprecated item 'std::heap::Heap': type renamed to `Global`
  --> src/main.rs:38:27
   |
38 |                 let ptr = Heap.realloc(self.ptr.as_ptr() as *mut _,
   |                           ^^^^

warning: use of deprecated item 'std::heap::Heap': type renamed to `Global`
  --> src/main.rs:47:29
   |
47 |                 Err(err) => Heap.oom(err),
   |                             ^^^^

warning: use of deprecated item 'std::heap::Heap': type renamed to `Global`
  --> src/main.rs:61:17
   |
61 |                 Heap.dealloc(self.ptr.as_ptr() as *mut _,
   |                 ^^^^

error[E0308]: mismatched types
  --> src/main.rs:38:40
   |
38 |                 let ptr = Heap.realloc(self.ptr.as_ptr() as *mut _,
   |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `std::ptr::NonNull`, found *-ptr
   |
   = note: expected type `std::ptr::NonNull<std::heap::Opaque>`
              found type `*mut _`

error[E0308]: mismatched types
  --> src/main.rs:40:40
   |
40 |                                        Layout::array::<T>(new_cap).unwrap());
   |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected usize, found struct `std::heap::Layout`
   |
   = note: expected type `usize`
              found type `std::heap::Layout`

error[E0599]: no method named `oom` found for type `std::heap::Global` in the current scope
  --> src/main.rs:47:34
   |
47 |                 Err(err) => Heap.oom(err),
   |                                  ^^^

error[E0308]: mismatched types
  --> src/main.rs:61:30
   |
[...]

PR #71, opened 2018-06-13
Fixed  of compilation on latest nightly locally, but still issue on playg…

izderadicka:
Fixing #70, works fine with local nightly, but but looks like very latest nightly has changed again.

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -28,7 +28,7 @@ For instance, a custom implementation of `Box` might write `Drop` like this:
 ```rust
 #![feature(ptr_internals, allocator_api, unique)]
 
-use std::alloc::{Global, GlobalAlloc, Layout};
+use std::alloc::{System, GlobalAlloc, Layout};
 use std::mem;
 use std::ptr::{drop_in_place, Unique};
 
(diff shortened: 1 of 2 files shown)


izderadicka:
Ok with latest update of nightly got same issue.

RalfJung:
Seems like https://github.com/rust-lang-nursery/nomicon/pull/73 did the same thing and hence broke this PR?

izderadicka:
@RalfJung  yes I think it's  almost the same except two things:
1. I think there is logical error in Drain as both Iterator and DoubleEndedIterator impl are using next_back from underlying iterator- so both going in same direction - backwards. 
2. I added some tests to vec-final.

Remainig diffs are just formating ( and I used System instead of Global but Global seems to be right name now).
If you're interested in two changes mentioned above let me know I'll  pull latest version and leave just these two.

RalfJung:
> I think there is logical error in Drain as both Iterator and DoubleEndedIterator impl are using next_back from underlying iterator- so both going in same direction - backwards.

That looks like a typo indeed but it was not introduced by #73. I'd prefer if @Gankro would take a look.^^

> I added some tests to vec-final.

I'm always in favor of tests. :D Do you know if `mdbook test` will run these tests?

Gankro:
CC @steveklabnik on mdbook

steveklabnik:
`mdbook test` runs `rustdoc --test` on each of your files, so it should.

RalfJung:
So @izderadicka would you like to make a new PR with your tests, and with the fix to `next_back` (that is likely needed to make the tests pass)?

izderadicka:
@RalfJung will look at it next week as I'm off this week. Closing this one.

PR #73, opened 2018-06-16
Fix nomicon for allocator changes

Mark-Simulacrum:
This doesn't update it to be idiomatic. I think @Gankro or someone else who is
more familiar with the allocator APIs should do that; it seems like `Unique<T>`
shouldn't be used anymore (it's marked as `doc(hidden)`) but I these fixes are
sufficient for an update for rust-lang/rust.

r? @steveklabnik


Mark-Simulacrum:
CI should now pass.

PR #74, opened 2018-06-20
document #[used]

japaric:
r? @oli-obk or @steveklabnik 

I'm not 100% sure about grouping all those features under an "ABI" section, but having a single #[used] section seemed overkill. I'm open to suggestions about grouping features differently, e.g. using a different section name.

Required by rust-lang/rust#51363

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -56,3 +56,5 @@
 	* [Final Code](vec-final.md)
 * [Implementing Arc and Mutex](arc-and-mutex.md)
 * [FFI](ffi.md)
+* [Application Binary Interface](abi.md)
+  * [#[used]](used.md)
--- a/src/used.md
+++ b/src/used.md
@@ -1,0 +1,66 @@
+## #[used]
+
+The #[used] attribute forces the *compiler* to keep a `static` variable in the output object file.
+This is useful for placing data at specific memory locations: for example, placing the vector table
+(interrupt table) in the memory location required by the ARM Cortex-M ABI: `0x0000_0000`.
+
+It's important to note that, on its own, `#[used]` has no effect on the behavior of the *linker*.
+That is the linker is free to drop a variable marked as `#[used]` when linking object files; thus,
+*a `#[used]` variable may not necessarily make it into the final binary / executable*.
+
+The guaranteed way to keep a variable in the final binary is to pair `#[used]` with the `EXTERN`
+linker script command. Linkers are lazy: once they have found all the symbols needed by the first /
+root object file they will stop looking at the other object files in their list of arguments.
+`EXTERN` forces the linker to look into the other object files until it finds the `EXTERN`-ed
+symbol.
+
+Below is an example that shows that both `#[used]` and `EXTERN` are required to keep `static`
+variables in an executable.
+
+``` rust
+#![feature(panic_implementation)]
+#![feature(used)]
+#![no_main]
+#![no_std]
+
+use core::panic::PanicInfo;
+
+// `#[no_mangle]` makes it easier to `EXTERN` this variable / symbol in the
+// linker script
+// `pub` is required to make this symbol *external*; the linker ignores
+// *internal* symbols when looking for an `EXTERN`-ed symbol
+#[no_mangle]
+#[used]
+pub static FOO: u32 = 0;
+
+// kept by the compiler, but dropped by the linker
+#[used]
+static BAR: u32 = 0;
+
+// dropped by the compiler
+#[allow(dead_code)]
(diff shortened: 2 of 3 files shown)


steveklabnik:
> I'm not 100% sure about grouping all those features under an "ABI" section, but having a single #[used] section seemed overkill. I'm open to suggestions about grouping features differently, e.g. using a different section name.

Hmm, yeah. I feel.. the same. Wonder if anyone else has thoughts.

> Required by rust-lang/rust#51363

FWIW, these days, only an issue needs to be opened, not a full PR, but I appreciate it! Makes it even easier on us.

src/used.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+``` console
+$ echo 'EXTERN(FOO);' > link.x
+
+$ rustc -O -C lto \
+    -C panic=abort -C relocation-model=static \
+    -C link-arg=-nostartfiles -C link-arg=-Wl,-Tlink.x \
+    --emit=link,obj \
+    foo.rs
```
Gankro: To maximize readability can you make this one `-C` per line?

src/used.md
```diff
@@ -0,0 +1,66 @@
...
+
+use core::panic::PanicInfo;
+
+// `#[no_mangle]` makes it easier to `EXTERN` this variable / symbol in the
+// linker script
+// `pub` is required to make this symbol *external*; the linker ignores
+// *internal* symbols when looking for an `EXTERN`-ed symbol
+#[no_mangle]
+#[used]
+pub static FOO: u32 = 0;
```
Gankro: I am of the soft opinion that foo/bar/baz are subpar names to use in example code. I'd prefer descriptive names like: USED_AND_EXTERN, JUST_USED, TOTALLY_DEAD

src/used.md
```diff
@@ -0,0 +1,66 @@
...
+static BAR: u32 = 0;
+
+// dropped by the compiler
+#[allow(dead_code)]
+static BAZ: u32 = 0;
+
+#[panic_implementation]
+fn panic(_: &PanicInfo) -> ! {
+    loop {}
+}
```
Gankro: This panic stuff seems like a big distraction, what's the deal? Just trying to make the smallest number of symbols possible? How bad is it if you don't do this?

src/used.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+``` console
```
Gankro: The text should ideally say what this block is. Linux?

src/used.md
```diff
@@ -0,0 +1,66 @@
...
+$ echo 'EXTERN(FOO);' > link.x
```
Gankro: pls comment this to explain what this is, something like "create a linker script for our EXTERN declarations".

src/used.md
```diff
@@ -0,0 +1,66 @@
...
+
+$ rustc -O -C lto \
+    -C panic=abort -C relocation-model=static \
+    -C link-arg=-nostartfiles -C link-arg=-Wl,-Tlink.x \
+    --emit=link,obj \
+    foo.rs
+
+$ nm -C foo.o
```
Gankro: I wouldn't expect a reader to have any idea what `nm` is, some exposition would be desirable.

src/used.md
```diff
@@ -0,0 +1,66 @@
...
+
+It's important to note that, on its own, `#[used]` has no effect on the behavior of the *linker*.
+That is the linker is free to drop a variable marked as `#[used]` when linking object files; thus,
+*a `#[used]` variable may not necessarily make it into the final binary / executable*.
+
+The guaranteed way to keep a variable in the final binary is to pair `#[used]` with the `EXTERN`
+linker script command. Linkers are lazy: once they have found all the symbols needed by the first /
+root object file they will stop looking at the other object files in their list of arguments.
+`EXTERN` forces the linker to look into the other object files until it finds the `EXTERN`-ed
+symbol.
```
Gankro: Is this universal? Only for *nix? Is this concept also applicable to windows?

src/SUMMARY.md
```diff
@@ -56,3 +56,5 @@
 	* [Final Code](vec-final.md)
 * [Implementing Arc and Mutex](arc-and-mutex.md)
 * [FFI](ffi.md)
+* [Application Binary Interface](abi.md)
```
Gankro: imo "application binary interface" is less a term than ABI is these days, so I'd call this section ABI (just as its neighbour is FFI). One of them there acronyms that's losts its meaning.

Gankro:
closing due to inactivity, sorry for missing you!

PR #77, opened 2018-07-07
Fix test failure, `oom` has been renamed to `handle_alloc_error`.

kennytm:
See rust-lang/rust#51543.

PR #79, opened 2018-07-16
Fix for Rust changes

RalfJung:
Also pin mdbook version until https://github.com/rust-lang-nursery/mdBook/issues/724 is resolved

### DIFF
--- a/src/vec-final.md
+++ b/src/vec-final.md
@@ -9,7 +9,7 @@ use std::ptr::{Unique, NonNull, self};
 use std::mem;
 use std::ops::{Deref, DerefMut};
 use std::marker::PhantomData;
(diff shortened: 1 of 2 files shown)


src/vec-final.md
```diff
@@ -47,7 +47,7 @@ impl<T> RawVec<T> {
 
             // If allocate or reallocate fail, oom
```
mati865: This PR missed comment modification, see https://github.com/rust-lang-nursery/nomicon/pull/77

RalfJung: Do you want to submit a comment fix?

mati865: I'm a bit time-constrained for now, there may be other places to check.

PR #82, opened 2018-08-08
Don’t recommend empty enums for opaque types

### LINKED ISSUE #29: FFI section suggests dangerous practice of using empty opaque type

RalfJung:
Supersedes https://github.com/rust-lang-nursery/nomicon/pull/44 (which seems to have gone stale).

Fixes #29.

### DIFF
--- a/src/ffi.md
+++ b/src/ffi.md
@@ -737,11 +737,11 @@ void foo(struct Foo *arg);
 void bar(struct Bar *arg);
 ```
 
-To do this in Rust, let’s create our own opaque types with `enum`:
+To do this in Rust, let’s create our own opaque types:
 
 ```rust
-pub enum Foo {}
-pub enum Bar {}
+#[repr(C)] pub struct Foo { private: [u8; 0] }
+#[repr(C)] pub struct Bar { private: [u8; 0] }
 
 extern "C" {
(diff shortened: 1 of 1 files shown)


gnzlbg:
LGTM. 

cc @ubsan @eddyb @steveklabnik

ubsan:
I'd rather have an explicit `opaque` type in the language, but in lieu of that, this is better than before.

gnzlbg:
@ubsan there is an already merged RFC for that (https://github.com/rust-lang/rust/issues/43467). This PR only updates the nomicon with better practices for the features that the language currently has. Once extern types are finished, this will need updating again.

ubsan:
@gnzlbg ooh! cool! I've never seen that! :D

PR #84, opened 2018-08-13
Remove `unique` feature which doesn't exist anymore

Bobo1239:
`Unique` was marked permanently-unstable in rust#46952 and its feature
flag was changed to `ptr_internals` which we're already using.

See https://github.com/rust-lang/rust/pull/46952.
Required to fix [toolstate](https://rust-lang-nursery.github.io/rust-toolstate/) as https://github.com/rust-lang/rust/pull/52644 introduced an error when an unknown feature is encountered. edit: This would also unblock nightlies.

### DIFF
--- a/src/destructors.md
+++ b/src/destructors.md
@@ -26,7 +26,7 @@ this is totally fine.
 For instance, a custom implementation of `Box` might write `Drop` like this:
 
 ```rust
-#![feature(ptr_internals, allocator_api, unique)]
+#![feature(ptr_internals, allocator_api)]
 
 use std::alloc::{Alloc, Global, GlobalAlloc, Layout};
 use std::mem;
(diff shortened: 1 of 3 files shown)


RalfJung:
> This would also unblock nightlies.

Nightlies are not blocked when the nomicon is broken. They are only blocked for rustfmt, clippy and rls.

Still, would be nice to get it green again!

RalfJung:
@Bobo1239 do you want to make a PR against rustc to update the nomicon submodule?

Bobo1239:
@RalfJung Sure, that was my initial intention anyways.
Regarding the nightlies: I somehow didn't see that rls was just fixed when I was checking so I assumed that nomicon was the blocker... :sweat_smile:

PR #85, opened 2018-08-13
Add fn to the list of invalid primitive values

mbrubeck:


rkruppe:
I don't think a dangling `fn` is necessarily a problem, and the motivation for dangling references being UB (we want to be able to speculate dereferences) doesn't apply to there. But I suppose there's no harm in being overly broad in the nomicon until an explicit decision has been made.

mbrubeck:
> I don't think a dangling `fn` is necessarily a problem

I thought it had to be forbidden because it could be called from safe code.

rkruppe:
That's a good point, handing a dangling `fn` to safe code is certainly a one-way ticket to UB. However, there's a distinction between invariants that can be assumed to hold by other code, and invariants where it's instant UB to ever produce a value that violates it. In this case, safe (or sound unsafe) code can certainy assume `fn`s are callable, but we may want to allow careful unsafe code to produce a dangling `fn` value if it's never called.

RalfJung:
What does it even mean for a `fn` to be "dangling"? Usually, the part of memory where the assembly is stored is not considered "valid" for access by the program.

> I thought it had to be forbidden because it could be called from safe code.

Safe code requires *way* more than we can make insta-UB. For example, safe code requires that a `fn` pointer will actually, when executed, not crash. We cannot even make this part of the definition of UB as that would make it cyclic.

So, we have to separate the discussion about invariants that must always be maintained lest you get UB, and the (much stronger) invariants that safe code can rely on. (Given how often I have written that in the last week, I should probably make this my next blog post.^^)

mbrubeck:
For now should I change this to forbid only null `fn` pointers?

RalfJung:
Yeah, I think that makes most sense. We certainly cannot have them be NULL.

rkruppe:
@RalfJung 
> What does it even mean for a fn to be "dangling"?

Not pointing to a function (of the right signature and ABI). Since we don't have a solution for code being unloaded at run time, that mostly just means making up an address that doesn't contain code and type-punning it to a `fn` value.


@nox 
> For now should I change this to forbid only null fn pointers?

I think either is fine. It's not that bad if the nomicon said something about dangling `fn`s that turned out to be wrong, and as you noted on IRC this text doesn't really make the distinction @RalfJung explained.

mbrubeck:
Updated.

PR #86, opened 2018-08-19
Test and fix to vec-final

izderadicka:
Fixed Iterator implementation for Drain
Added some tests to code

@RalfJung  - as discussed in #71 I've created separate PR for these two changes
@steveklabnik  -  in #71 you noted that test should run during `mdbook test`, but it does not seem to be so (if I modify a test to fail nothing happens) -  is there anything else needed for tests to run?

### DIFF
--- a/src/vec-final.md
+++ b/src/vec-final.md
@@ -304,76 +304,76 @@ impl<'a, T> Drop for Drain<'a, T> {
     }
 }
 
-# fn main() {}
+# fn main() {
+#     tests::create_push_pop();
+#     tests::iter_test();
+#     tests::test_drain();
+#     tests::test_zst();
+# }
 
-#[cfg(test)]
-mod tests {
-    use super::*;
-    #[test]
-    fn create_push_pop() {
-        let mut v = Vec::new();
-        v.push(1);
-        assert_eq!(1, v.len());
-        assert_eq!(1, v[0]);
-        for i in v.iter_mut() {
-            *i += 1;
-        }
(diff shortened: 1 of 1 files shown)


izderadicka:
@steveklabnik - I think I get it with tests -  mdbook is running doctests only - so I modify code bit to run tests as doctests - also tests are now hidden, as main function is hidden too,  so they do not appear in book - to make final example clearer.

src/vec-final.md
```diff
@@ -304,5 +304,77 @@ impl<'a, T> Drop for Drain<'a, T> {
...
+#     use super::*;
+#     pub fn create_push_pop() {
+#         let mut v = Vec::new();
+#         v.push(1);
+#         assert_eq!(1, v.len());
+#         assert_eq!(1, v[0]);
+#         for i in v.iter_mut() {
+#             *i += 1;
+#         }
+#         v.insert(0, 1);
```
RalfJung: Maybe insert a different value here so we can distinguish them, making sure the right one gets indeed popped?

izderadicka:
@RalfJung - Yes test are running now - as doctest is running `main` and main is running our tests. Also tested that if assertion fails then `mdbook test` fails as the test panics.

RalfJung:
LGTM except for changing `v.insert(0, 1);` to `v.insert(0, 2);` (or any value other than `1` and `10`).

izderadicka:
@RalfJung - changed test as required

PR #94, opened 2018-10-23
There are now 3 unsafe traits in rust

vorner:
Updating the stale information, adding the GlobalAllocator trait.

To be honest, I'm not 100% sure there are *just* these three, but I tried grepping the standard library source code for `unsafe trait` and all the other matches seem to be still unstable.

To give a bigger context to the pull request, I feel like the nomicon would deserve some more work and love. So I'd like to start by going through it and at least updating obviously outdated information and filling in TODOs. If it's OK with everyone, I would like to be sending such pull request for a while (likely in smallish chunks, as time allows).

### DIFF
--- a/src/safe-unsafe-meaning.md
+++ b/src/safe-unsafe-meaning.md
@@ -41,12 +41,15 @@ The standard library has a number of unsafe functions, including:
 * All FFI (Foreign Function Interface) functions are `unsafe` to call because the
   other language can do arbitrary operations that the Rust compiler can't check.
 
-As of Rust 1.0 there are exactly two unsafe traits:
+As of Rust 1.29.2 the standard library defines the following unsafe traits
+(there are others, but they are not stabilized yet and some of them may never
+be):
 
-* `Send` is a marker trait (a trait with no API) that promises implementors are
-  safe to send (move) to another thread.
-* `Sync` is a marker trait that promises threads can safely share implementors
+* [`Send`] is a marker trait (a trait with no API) that promises implementors
+  are safe to send (move) to another thread.
+* [`Sync`] is a marker trait that promises threads can safely share implementors
   through a shared reference.
+* [`GlobalAlloc`] allows customizing the memory allocator of the whole program.
 
 Much of the Rust standard library also uses Unsafe Rust internally. These
 implementations have generally been rigorously manually checked, so the Safe Rust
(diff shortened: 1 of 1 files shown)


src/safe-unsafe-meaning.md
```diff
@@ -123,23 +125,31 @@ Rust has traditionally avoided doing this because it makes Unsafe
...
-`Ord` implementation. The decision of whether to mark your own traits `unsafe`
-depends on the same sort of consideration. If `unsafe` code can't reasonably
-expect to defend against a bad implementation of the trait, then marking the
-trait `unsafe` is a reasonable choice.
+`Ord` implementation. Similarly, `GlobalAllocator` is keeping accounts of all
+the memory in the program and nobody can hope to deal with things like the
+allocator serving two different requests with the same chunk of memory. The
+decision of whether to mark your own traits `unsafe` depends on the same sort of
+consideration. If `unsafe` code can't reasonably expect to defend against a bad
+implementation of the trait, then marking the trait `unsafe` is a reasonable
```
RalfJung: Might be worth expanding on "bad". (I know the original text didn't do that, so feel free to ignore this comment if you prefer.) Concretely, the point is that we expect the trait to follow more rules than the types suggest (this applies to `Ord` and `GlobalAllocator` alike), and that unsafe code can *rely* on the trait following these rules. The latter part is mentioned but I feel it is worth saying that there is a base promise made for every trait ("the methods act safety at their given type"), and `unsafe` merely says that we require more.

RalfJung:
> To give a bigger context to the pull request, I feel like the nomicon would deserve some more work and love. So I'd like to start by going through it and at least updating obviously outdated information and filling in TODOs. If it's OK with everyone, I would like to be sending such pull request for a while (likely in smallish chunks, as time allows).

Sounds great! I'd be happy to help reviewing your PRs. However, I do not have r+ powers in this repo.

vorner:
OK, I'll try some rewording (but I don't know if I'll get to it today).

vorner:
Maybe it's more clear now?

src/safe-unsafe-meaning.md
```diff
@@ -41,12 +41,14 @@ The standard library has a number of unsafe functions, including:
 * All FFI (Foreign Function Interface) functions are `unsafe` to call because the
   other language can do arbitrary operations that the Rust compiler can't check.
 
-As of Rust 1.0 there are exactly two unsafe traits:
+As of Rust 1.29.2 there these stable unsafe traits (there are others, but they
```
Gankro: trying out this suggesting thing for the first time??

```suggestion
As of Rust 1.29.2 the standard library defines the following unsafe traits (there are others, but they
```

src/safe-unsafe-meaning.md
```diff
@@ -41,12 +41,14 @@ The standard library has a number of unsafe functions, including:
...
+are not stabilized yet and some of them might never be):
```
Gankro:
```suggestion
are not stabilized yet and some of them may never be):
```

src/safe-unsafe-meaning.md
Gankro: As an aside the nomicon is not strictly bound to documenting stable things; indeed its early focus involved many unstable things.

src/safe-unsafe-meaning.md
```diff
@@ -122,24 +124,35 @@ The decision of whether to mark a trait `unsafe` is an API design choice.
...
+the pervasive unsafety of making these two traits `unsafe`. And not many people
+are going to *implement* memory allocators (not even use directly, for that
+matter).
 
 This is the balance between Safe and Unsafe Rust. The separation is designed to
 make using Safe Rust as ergonomic as possible, but requires extra effort and
 care when writing Unsafe Rust. The rest of this book is largely a discussion
 of the sort of care that must be taken, and what contracts Unsafe Rust must uphold.
 
+[`Send`]: https://doc.rust-lang.org/std/marker/trait.Send.html
```
Gankro: Makes this point to the same version of the stdlib docs that the visible copy of the nomicon does (stable -> stable, nightly -> nightly, etc)

```suggestion
[`Send`]: ../std/marker/trait.Send.html
```

src/safe-unsafe-meaning.md
```diff
@@ -122,24 +124,35 @@ The decision of whether to mark a trait `unsafe` is an API design choice.
...
+[`Sync`]: https://doc.rust-lang.org/std/marker/trait.Sync.html
```
Gankro:
```suggestion
[`Sync`]: ../std/marker/trait.Sync.html
```

src/safe-unsafe-meaning.md
```diff
@@ -122,24 +124,35 @@ The decision of whether to mark a trait `unsafe` is an API design choice.
...
+[`GlobalAlloc`]: https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html
```
Gankro:
```suggestion
[`GlobalAlloc`]: ../std/alloc/trait.GlobalAlloc.html
```

src/safe-unsafe-meaning.md
```diff
@@ -122,24 +124,35 @@ The decision of whether to mark a trait `unsafe` is an API design choice.
...
+a reasonable choice.
 
 As an aside, while `Send` and `Sync` are `unsafe` traits, they are *also*
 automatically implemented for types when such derivations are provably safe
 to do. `Send` is automatically derived for all types composed only of values
 whose types also implement `Send`. `Sync` is automatically derived for all
 types composed only of values whose types also implement `Sync`. This minimizes
-the pervasive unsafety of making these two traits `unsafe`.
+the pervasive unsafety of making these two traits `unsafe`. And not many people
+are going to *implement* memory allocators (not even use directly, for that
```
Gankro:
```suggestion
are going to *implement* memory allocators (or use them directly, for that
```

src/safe-unsafe-meaning.md
vorner: I noticed there are some unsafe things, but I thought it was more like out of necessity. Anyway, there's a bit too many of the unstable unsafe traits. I'm afraid the full list would look a bit scary.

Gankro:
@steveklabnik what's our policy on how to handle/squash these suggested change commits

steveklabnik:
We don’t really *require* anything, though a squash and rebase before an r+ is always nice, if things eventually need to be reverted it makes it so much easier.

PR #101, opened 2018-11-04
Lifetimes: Updates to incorporate NLL

vorner:
* Updated the explanations around lifetimes a bit.
* Made sure the examples that should fail still fail in edition 2018.
* Prefer `rust,compile_fail` instead of `rust,ignore` ‒ the latter
  allows the user to click on button and see the actual compile errors.
  Also, this'll tell us if something stops failing.

### DIFF
--- a/src/lifetimes.md
+++ b/src/lifetimes.md
@@ -1,9 +1,17 @@
 # Lifetimes
 
-Rust enforces these rules through *lifetimes*. Lifetimes are effectively
-just names for scopes somewhere in the program. Each reference,
-and anything that contains a reference, is tagged with a lifetime specifying
-the scope it's valid for.
+Rust enforces these rules through *lifetimes*. Lifetimes are named
+regions of code that a reference must be valid for. Those regions
+may be fairly complex, as they correspond to paths of execution
+in the program. There may even be holes in these paths of execution,
+as it's possible to invalidate a reference as long as it's reinitialized
+before it's used again. Types which contain references (or pretend to)
+may also be tagged with lifetimes so that Rust can prevent them from
+being invalidated as well.
+
+In most of our examples, the lifetimes will coincide with scopes. This is
+because our examples are simple. The more complex cases where they don't
+coincide are described below.
 
 Within a function body, Rust generally doesn't let you explicitly name the
 lifetimes involved. This is because it's generally not really necessary
@@ -23,10 +31,10 @@ syrup even -- around scopes and lifetimes, because writing everything out
 explicitly is *extremely noisy*. All Rust code relies on aggressive inference
 and elision of "obvious" things.
 
-One particularly interesting piece of sugar is that each `let` statement implicitly
-introduces a scope. For the most part, this doesn't really matter. However it
-does matter for variables that refer to each other. As a simple example, let's
-completely desugar this simple piece of Rust code:
+One particularly interesting piece of sugar is that each `let` statement
+implicitly introduces a scope. For the most part, this doesn't really matter.
+However it does matter for variables that refer to each other. As a simple
+example, let's completely desugar this simple piece of Rust code:
 
 ```rust
 let x = 0;
@@ -85,7 +93,7 @@ z = y;
 
 Alright, let's look at some of those examples from before:
 
-```rust,ignore
+```rust,compile_fail
 fn as_str(data: &u32) -> &str {
     let s = format!("{}", data);
     &s
@@ -169,7 +177,7 @@ our implementation *just a bit*.)
 
 How about the other example:
 
-```rust,ignore
+```rust,compile_fail
 let mut data = vec![1, 2, 3];
 let x = &data[0];
 data.push(4);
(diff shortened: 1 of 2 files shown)


RalfJung:
>  Prefer rust,compile_fail instead of rust,ignore ‒ the latter
allows the user to click on button and see the actual compile errors.
Also, this'll tell us if something stops failing.

You mean "the *former*"?

src/lifetimes.md
```diff
@@ -213,3 +218,82 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+
+
+
+# The area covered by a lifetime
+
+The lifetime (sometimes called a *borrow*) is *alive* from the place it is
+created to its last use. The borrowed thing needs to outlive only borrows that
+are alive. This looks simple, but there are few subtleties.
+
+This compiles, because after printing `x`, it is no longer needed, so it doesn't
```
RalfJung: This "this" is quite far away from the example it refers to, so I was confused for a bit here. Could you move it closer, or else make the reference more explicit?

vorner: I moved the paragraph below that to the very end, that might be a better place.

vorner:
> You mean "the former"?

Hmm, yes. I'll rewrite that on the before-merge rebase.

Gankro:
note I intend to review this soon but I want to touch base with the lang team to understand NLL better before going over this in detail

vorner:
Hello. Any progress here?

vorner:
Hello

@Gankro Is there a realistic chance you'd get around to it, or should we try to seek some other way of validating it that doesn't block on you? It's sitting here without progress for 5 months and the book contains outdated information that is no longer true :-(.

Maybe asking someone from the NLL implementers to skim the text for blatant factual lies and and you having a look if the text fits nicely with the rest of the book and is understandable would work?

Gankro:
Yeah reviewing this has been my white whale. A true disaster I have created and inflicted upon you. :/

Part of the issue is that this substantial of a change requires actual, like, editing for effective communication, which I find very uncomfortable to do? I would definitely appreciate technical feedback from domain experts here, but this also needs the attention of like, a Writing Professional.

src/lifetimes.md
```diff
@@ -213,3 +218,82 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+data.push(4);
+```
+
+However, if the value has a destructor, the destructor is run at the end of the
+scope. And running the destructor is considered a use ‒ obviously the last one.
+So, this will *not* compile.
+
+```rust,edition2018,compile_fail
+#[derive(Debug)]
+struct X<'a>(&'a i32);
```
Gankro: clarity nit: Can you rename this to Borrow?

src/lifetimes.md
```diff
@@ -1,9 +1,14 @@
 # Lifetimes
 
 Rust enforces these rules through *lifetimes*. Lifetimes are effectively
-just names for scopes somewhere in the program. Each reference,
-and anything that contains a reference, is tagged with a lifetime specifying
-the scope it's valid for.
+just names for areas of code somewhere in the program (you could also look at
+them as time intervals, except that the beginning and end of the time is denoted
+by when a line of code executes). Each reference, and anything that contains a
+reference, is tagged with a lifetime specifying the time it's valid for.
```
Gankro: I'm not sure I like this time metaphor. A more apt one might be "paths of execution"?

src/lifetimes.md
```diff
@@ -213,3 +218,82 @@ totally ok*, because it keeps us from spending all day explaining our program
...
+
+impl Drop for X<'_> {
+    fn drop(&mut self) {}
+}
+
+let mut data = vec![1, 2, 3];
+let x = X(&data[0]);
+println!("{:?}", x);
+data.push(4);
+// Here, the destructor is run and therefore this'll fail to compile.
```
Gankro:
```suggestion
// Here, the destructor is run and prevents the borrow from expiring
```

vorner:
I've applied the suggestions into a new fixup. If it looks good, I'll squash the fixups to have nicer history.

I'm afraid this is based on old version of master that no longer compiles all the snippets (in other chapters). I don't want to mess the review with a full rebase yet, though, so I'm postponing the rebase to current master for when the changes here are ACKed. Is that OK?

src/lifetimes.md
```diff
@@ -1,9 +1,15 @@
...
 
 Rust enforces these rules through *lifetimes*. Lifetimes are effectively
-just names for scopes somewhere in the program. Each reference,
-and anything that contains a reference, is tagged with a lifetime specifying
-the scope it's valid for.
+just names for areas of code somewhere in the program (you could also look at
+them as time intervals, except that the beginning and end of the time is denoted
+by when a line of code executes). Each reference, and anything that contains a
+reference, is tagged with a lifetime specifying the paths of executions it's
+valid for.
```
Gankro: This paragraph is now a bit too much of a mess of metaphors. Suggested rewrite:

```markdown
Rust enforces these rules through *lifetimes*. Lifetimes are named
regions of code that a reference must be valid for. Those regions
may be fairly complex, as they correspond to paths of execution
in the program. There may even be holes in these paths of execution, 
as it's possible to invalidate a reference as long as it's reinitialized
before it's used again. Types which contain references (or pretend to)
may also be tagged with lifetimes so that Rust can prevent them from
being invalidated as well.
```

vorner: Not the first time that happened to me, I guess my brain just works mostly on metaphors most of the time :-|. Anyway, adopting your version, thank you.

Gankro:
ready to land?

vorner:
Yes :-).

PR #102, opened 2018-11-09
Subtyping rewrite

Gankro:
[Rendered](https://github.com/Gankro/nomicon/blob/subtyping2/src/subtyping.md)

This is more or less a complete rewrite of the section, as an alternative to #99

### DIFF
--- a/src/subtyping.md
+++ b/src/subtyping.md
@@ -168,17 +168,17 @@ in Rust, though it does in fact exist.
 Here is a table of important variances which the rest of this section will be devoted
 to trying to explain:
 
-|   |               |     'a    |         T         |     U     |
-|---|---------------|:---------:|:-----------------:|:---------:|
-| * | &'a T         | covariant | covariant         |           |
-| * | &'a mut T     | covariant | invariant         |           |
-| * | Box<T>        |           | covariant         |           |
-|   | Vec<T>        |           | covariant         |           |
-| * | UnsafeCell<T> |           | invariant         |           |
-|   | Cell<T>       |           | invariant         |           |
-| * | fn(T) -> U    |           | **contra**variant | covariant |
-|   | *const T      |           | covariant         |           |
-|   | *mut T        |           | invariant         |           |
+|   |                 |     'a    |         T         |     U     |
+|---|-----------------|:---------:|:-----------------:|:---------:|
+| * | `&'a T `        | covariant | covariant         |           |
+| * | `&'a mut T`     | covariant | invariant         |           |
+| * | `Box<T>`        |           | covariant         |           |
+|   | `Vec<T>`        |           | covariant         |           |
+| * | `UnsafeCell<T>` |           | invariant         |           |
+|   | `Cell<T>`       |           | invariant         |           |
+| * | `fn(T) -> U`    |           | **contra**variant | covariant |
+|   | `*const T`      |           | covariant         |           |
+|   | `*mut T`        |           | invariant         |           |
 
 The types with \*'s are the ones we will be focusing on, as they are in
 some sense "fundamental". All the others can be understood by analogy to the others:
--- a/src/subtyping.md
+++ b/src/subtyping.md
@@ -3,15 +3,15 @@
 Subtyping is a relationship between types that allows statically typed
 languages to be a bit more flexible and permissive.
 
-Subtyping in Rust is a bit different from subtyping in other languages. This leads
-to examples of subtyping being *a bit* convoluted. This is especially unforunate
-because subtyping, and especially variance, are actually really hard to
-understand properly. As in, even compiler writers mess it up all the time.
+Subtyping in Rust is a bit different from subtyping in other languages. This
+makes it harder to give simple examples, which is a problem since subtyping,
+and especially variance, are already hard to understand properly. As in,
+even compiler writers mess it up all the time.
 
-In order to make examples that are simple and concise, this section will consider a
-*small* extension to the Rust language to introduce subtyping in a way that is more
-similar to other languages. After establishing concepts and issues under this simpler
-system, we will then relate it back to how subtyping actually occurs in Rust.
+To keep things simple, this section will consider a small extension to the
+Rust language that adds a new and simpler subtyping relationship. After
+establishing concepts and issues under this simpler system,
+we will then relate it back to how subtyping actually occurs in Rust.
 
 So here's our simple extension, *Objective Rust*, featuring three new types:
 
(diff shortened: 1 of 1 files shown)


src/subtyping.md
```diff
@@ -59,200 +139,284 @@ To handle it, we need to talk about *variance*.
 Variance is where things get a bit complicated.
 
 Variance is a property that *type constructors* have with respect to their
-arguments. A type constructor in Rust is a generic type with unbound arguments.
-For instance `Vec` is a type constructor that takes a `T` and returns a
-`Vec<T>`. `&` and `&mut` are type constructors that take two inputs: a
+arguments. A type constructor in Rust is any generic type with unbound arguments.
+For instance `Vec` is a type constructor that takes a type `X` and returns
+`Vec<X>`. `&` and `&mut` are type constructors that take two inputs: a
```
Havvy: Why `X` and not `T`?

src/subtyping.md
```diff
@@ -59,200 +139,284 @@ To handle it, we need to talk about *variance*.
...
-* `UnsafeCell<T>`, `Cell<T>`, `RefCell<T>`, `Mutex<T>` and all other
-  interior mutability types are invariant over T (as is `*mut T` by metaphor)
+The problem with making `&mut T` covariant over `T` is that it gives us the power
+to modify the original value *when we don't remember all of its constraints*.
+And so, we can make someone have a Dog when they're certain they still have a Cat.
 
-To understand why these variances are correct and desirable, we will consider
-several examples.
+With that established, we can easily see why `&T` being covariant over `T` *is*
+sound: it doesn't let you modify the value, only look at it. Without anyway to
```
RalfJung: "Without any way"

src/subtyping.md
```diff
@@ -59,200 +139,284 @@ To handle it, we need to talk about *variance*.
...
+mutate, there's no way for us to mess with any details. We can also see why
+UnsafeCell and all the other interior mutability types must be invariant: they
```
RalfJung:
```suggestion
`UnsafeCell` and all the other interior mutability types must be invariant: they
```

src/subtyping.md
```diff
@@ -59,200 +139,284 @@ To handle it, we need to talk about *variance*.
...
+make `&T` work like `&mut T`!
 
-We have already covered why `&'a T` should be covariant over `'a` when
-introducing subtyping: it's desirable to be able to pass longer-lived things
-where shorter-lived things are needed.
+Now what about the lifetime on references? Why is it ok for both kinds of references
+to be covariant over their lifetimes. Well, here's a two-pronged argument:
```
RalfJung:
```suggestion
to be covariant over their lifetimes? Well, here's a two-pronged argument:
```

src/subtyping.md
```diff
@@ -59,200 +139,284 @@ To handle it, we need to talk about *variance*.
...
+of subtyping in Rust*. The only reason we have subtyping is so we can pass
+long-lived things where short-lived things are expected. So it better work!
 
-However this logic doesn't apply to `&mut`. To see why `&mut` should
-be invariant over T, consider the following code:
+Second, and more seriously, references *own* their lifetimes, while they are
+only borrowing their referents. If you shrink down a reference's lifetime when
+you hand it to someone, that location now has a reference which owns the smaller
+lifetime. There's no way to mess with original reference's lifetime using the
+other one.
```
RalfJung: I find this use of "own" somewhat hard to follow.

The way I think about it: The lifetime in a reference is just a *lower bound*. It is how long we are *sure* the thing is there, but it may be there longer and we wouldn't care. That's why making it shorter is always okay. Do you think that would work as an explanation?

src/subtyping.md
```diff
@@ -59,200 +139,284 @@ To handle it, we need to talk about *variance*.
...
+This is why we get the error when we try to assign `&spike` to `spike_str`. The
+compiler has worked backwards to conclude `spike_str` must live forever, and `&spike`
+simply can't live that long.
+
+So even though references are covariant over their lifetimes, they "inherit" invariance
+whenever they're put into a context that could do something bad with that. In this case,
+we inherited invariance as soon as we put our reference inside an `&mut T`.
+
+As it turns out, the argument for why it's ok for Box (and Vec, Hashmap, etc.) to
+be covariant is pretty similar to the argument for why it's ok for why it's ok for
```
RalfJung:
```suggestion
be covariant is pretty similar to the argument for why it's ok for
```

src/subtyping.md
Gankro: I tried rewording it a bit, wdyt?

src/subtyping.md
```diff
@@ -59,200 +141,288 @@ To handle it, we need to talk about *variance*.
...
+in Rust, though it does in fact exist.
+
+Here is a table of important variances which the rest of this section will be devoted
+to trying to explain:
+
+|   |               |     'a    |         T         |     U     |
+|---|---------------|:---------:|:-----------------:|:---------:|
+| * | &'a T         | covariant | covariant         |           |
+| * | &'a mut T     | covariant | invariant         |           |
+| * | Box<T>        |           | covariant         |           |
```
dtolnay: Please use `Box\<T\>` or ``` `Box<T>` ``` so that the T appears in the rendered page.

dtolnay:
Trait objects are not covered. In my experience a trait object `Trait<'a, T>` is always invariant in `'a` and `T`. This is probably the biggest point of friction where variance doesn't just do the right thing, so might be a common reason that people arrive at this documentation. There is currently no way for a trait to require impls of itself to have a particular variance.

[In Scala](https://www.scala-lang.org/api/current/scala/Function1.html) type parameters of traits can require a variance:

```scala
trait Function1[-T1, +R] extends AnyRef
```

Whereas in Rust:

```rust
fn fn_ptr<'a>(f: &fn(&'a u8)) {
    // okay
    let _: &fn(&'static u8) = f;
}

fn fn_trait_object<'a>(f: &dyn Fn(&'a u8)) {
    // error because the trait object is invariant in the type of argument 0
    let _: &dyn Fn(&'static u8) = f;
}
```

dtolnay:
Maybe also discuss the role of PhantomData in the context of the trouble with unused type parameters, which is mainly about variance (as I understand it). We need it to deduce the variance with respect to `T`.

```rust
struct S<T> {}
```

```console
error[E0392]: parameter `T` is never used
 --> src/main.rs:1:10
  |
1 | struct S<T> {}
  |          ^ unused type parameter
  |
  = help: consider removing `T` or using a marker such as `std::marker::PhantomData`
```

RalfJung:
> Trait objects are not covered. In my experience a trait object Trait<'a, T> is always invariant in 'a and T. This is probably the biggest point of friction where variance doesn't just do the right thing, so might be a common reason that people arrive at this documentation. There is currently no way for a trait to require impls of itself to have a particular variance.

Ah, good point, I had never thought about that. However, to even talk about variance in trait objects we would have to first have a notion of a "trait constructor" that takes parameters and defines a trait -- and you couldn't even do something like `impl Trait for &'static i32` unless `Trait` was invariant, you'd have to always implement such a trait in a generic way such that the same instance still applies after applying variance. All of that seems non-trivial to me. I am actually pretty happy Rust's subtyping is so limited; I was told one of the main reasons it took a decade to prove anything formal about Scala is that it's type system is crazy complicated due to subtyping.

Your example is a bit contrived anyway, we'd usually write `fn fn_trait_object(f: &dyn for<'a> Fn(&'a u8))` or, shorter, `fn fn_trait_object(f: &dyn Fn(&u8))`. Then using this with `'static` would work.

Gankro:
PhantomData is the next-next page, after we've discussed the drop check, so that we have full context for its existence.

Mentioning that subtyping *doesn't* work with dyn traits is a good idea though, I'll think about how to incorporate it.

RalfJung:
Variance inference however does seem like an important topic, and AFAIK it is the sole reason why we do not accept unused type parameters.  Seems relevant to be discussed here? And once you discuss variance inference you kind of have to mention `PhantomData`.

The fact that `PhantomData` *also* affects dropck seems orthogonal to me.

PR #104, opened 2018-11-17
A little note that HRTBs can be used for types too

vorner:


RalfJung:
But is that correct? AFAIK `for<T>` is not allowed.

vorner:
Hmm, you are probably right. Yesterday I was *sure* I've used it somewhere, but now I can't seem to find the piece of code and trivial examples give „only lifetime parameters can be used in this context“.

I'll watch out for the use I must have done and then find out in what context these are allowed.

PR #106, opened 2018-11-21
cleanups for Rust 2018

Gankro:
round 1, covering the first two chapters.

### DIFF
--- a/src/exotic-sizes.md
+++ b/src/exotic-sizes.md
@@ -14,7 +14,7 @@ known size or alignment. On the surface, this is a bit nonsensical: Rust *must*
 know the size and alignment of something in order to correctly work with it! In
 this regard, DSTs are not normal types. Because they lack a statically known
 size, these types can only exist behind a pointer. Any pointer to a
-DST consequently becomes a *fat* pointer consisting of the pointer and the
+DST consequently becomes a *wide* pointer consisting of the pointer and the
 information that "completes" them (more on this below).
 
 There are two major DSTs exposed by the language:
@@ -66,6 +66,10 @@ fn main() {
     println!("{} {:?}", dynamic.info, &dynamic.data);
 }
 ```
+
+(Yes, custom DSTs are a largely half-baked feature for now.)
+
+
 
 
 
@@ -157,11 +161,16 @@ because that wouldn't make sense.
 
 We recommend against modelling C's `void*` type with `*const Void`.
 A lot of people started doing that but quickly ran into trouble because
-people want to convert raw pointers to references when doing FFI, and making an
-`&Void` is Undefined Behaviour. `*const ()` (or equivalent) works just as well,
-and can be made into a reference without any safety problems. The only downside
-of modeling `void*` with `*const ()` is that attempts to read or write values
-will silently succeed (doing nothing), instead of producing a compiler error.
+Rust doesn't really have any safety guards against trying to instantiate
+empty types with unsafe code, and if you do it, it's Undefined Behaviour.
+This was especially problematic because developers had a habit of converting
+raw pointers to references and `&Void` is *also* Undefined Behaviour to
+construct.
+
+`*const ()` (or equivalent) works reasonably well for `void*`, and can be made
+into a reference without any safety problems. It still doesn't prevent you from
+trying to read or write values, but at least it compiles to a no-op instead
+of UB.
 
 
 
--- a/src/exotic-sizes.md
+++ b/src/exotic-sizes.md
@@ -1,7 +1,7 @@
 # Exotically Sized Types
 
-Most of the time, we think in terms of types with a fixed, positive size. This
-is not always the case, however.
+Most of the time, we expect types to have a statically known and positive size.
+This isn't always the case in Rust.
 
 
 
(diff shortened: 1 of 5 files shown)


src/exotic-sizes.md
```diff
@@ -97,7 +125,7 @@ enum Void {} // No variants = EMPTY
 ```
 
 Empty types are even more marginal than ZSTs. The primary motivating example for
-Void types is type-level unreachability. For instance, suppose an API needs to
+an empty type is type-level unreachability. For instance, suppose an API needs to
 return a Result in general, but a specific case actually is infallible. It's
 actually possible to communicate this at the type level by returning a
 `Result<T, Void>`. Consumers of the API can confidently unwrap such a Result
```
ubsan: Should probably change this to `Result<T, !>`, as the standard empty type.

src/other-reprs.md
```diff
@@ -15,18 +15,26 @@ or C++. Any type you expect to pass through an FFI boundary should have
...
 `repr(C)` can be applied to types that will be nonsensical or problematic if
 passed through the FFI boundary.
 
 * ZSTs are still zero-sized, even though this is not a standard behavior in
 C, and is explicitly contrary to the behavior of an empty type in C++, which
-still consumes a byte of space.
+says they should still consume a byte of space.
 
-* DST pointers (fat pointers), tuples, and enums with fields are not a concept
+* DST pointers (fat pointers) and tuples are not a concept
```
ubsan: I would prefer if you said either `thick pointers` or `fat/thick pointers`. I'm, at least personally, attempting to change this language right now.

src/repr-rust.md
```diff
@@ -2,12 +2,12 @@
...
 alignment of a type specifies what addresses are valid to store the value at. A
-value of alignment `n` must only be stored at an address that is a multiple of
+value with alignment `n` must only be stored at an address that is a multiple of
 `n`. So alignment 2 means you must be stored at an even address, and 1 means
 that you can be stored anywhere. Alignment is at least 1, and always a power
 of 2. Most primitives are generally aligned to their size, although this is
-platform-specific behavior. In particular, on x86 `u64` and `f64` may be only
-aligned to 32 bits.
+platform-specific behavior. For example, on x86 `u64` and `f64` may be only
+aligned to 4 (32 bits).
```
ubsan: probably want to write `4 bytes (32 bits)`

src/exotic-sizes.md
```diff
@@ -1,69 +1,98 @@
...
-struct Foo {
+struct MySuperSlice {
     info: u32,
     data: [u8],
 }
 ```
 
+Although such a type is largely useless without a way to construct it. Currently the
+only properly supported way to create a custom DST is by making your type generic
+and performing an *unsizing coercion*:
```
steveklabnik: i have wanted to know this, A+ adding it in here

src/other-reprs.md
```diff
@@ -15,18 +15,26 @@ or C++. Any type you expect to pass through an FFI boundary should have
 necessary to soundly do more elaborate tricks with data layout such as
 reinterpreting values as a different type.
 
-However, the interaction with Rust's more exotic data layout features must be
+We strongly recommend using [rust-bindgen][] and/or [cbdingen][] to manage your FFI
```
steveklabnik: markdown ProTip: you don't need the empty `[]`s

src/other-reprs.md
steveklabnik: i also have been using "double pointer" instead

src/exotic-sizes.md
```diff
@@ -125,9 +153,30 @@ But this trick doesn't work yet.
...
 actually valid to construct, but dereferencing them is Undefined Behavior
-because that doesn't actually make sense. That is, you could model C's `void *`
-type with `*const Void`, but this doesn't necessarily gain anything over using
-e.g. `*const ()`, which *is* safe to randomly dereference.
+because that wouldn't make sense.
+
+We recommend against modelling C's `void*` type with `*const Void`.
+A lot of people started doing that but quickly ran into trouble because
+people want to convert raw pointers to references when doing FFI, and making an
+`&Void` is Undefined Behaviour. `*const ()` (or equivalent) works just as well,
```
RalfJung: Well, is it? It might be, it might not be. It is certainly a footgun though.

src/exotic-sizes.md
```diff
@@ -125,9 +153,30 @@ But this trick doesn't work yet.
...
+and can be made into a reference without any safety problems. The only downside
+of modeling `void*` with `*const ()` is that attempts to read or write values
+will silently succeed (doing nothing), instead of producing a compiler error.
```
RalfJung: Reading an `*const Void` also doesn't necessarily produce an error -- instead of nothing, you just got UB.

src/other-reprs.md
RalfJung: There are no two pointers in a slice though, so "double" seems like a bad choice to me.

I don't know why you want to change language here, but shouldn't that also be changed in the [API docs then](https://doc.rust-lang.org/nightly/std/ops/trait.CoerceUnsized.html)? Also I am concerned about understandability, "fat pointer" is a standard term (even used in academia for "pointers annotated with stuff") that many will have heard before, using a different term puts up an additional unnecessary barrier. But more generally I wonder if there has been discussion about this where I could read up on your reasoning?

steveklabnik: short for "double-width pointer"; all the current ones are two usizes, even if the second isn't a pointer itself.

src/other-reprs.md
Gankro: You can leave them out, but using them makes the content slightly more robust and unambiguous, similar to quoting keys in JS

src/exotic-sizes.md
Gankro: The rust compiler literally emits a warning that your code is unreachable, and then your program SIGILL's.

src/exotic-sizes.md
Gankro: ! is still unstable and trapped in bikeshed hell, so I don't really have a strong motivation to mention it

src/exotic-sizes.md
ubsan: I feel like this is a bug in rustc...

RalfJung: How are you making the `&Void`?

Gankro: https://play.rust-lang.org/?version=nightly&mode=release&edition=2018&gist=23236a60b8801319313ff8d475fdbb0e

MIRI also aborts for hitting unreachable code

Gankro: I think it's safe to conservatively call this UB in the nomicon and then soften the language if y'all ever settle on something more precise/subtle

RalfJung: Yeah well you are constructing a `!` here and taking its address. Of course that does not work, and never will.

I thought of something more like https://play.rust-lang.org/?version=nightly&mode=release&edition=2018&gist=d4002e1c51c73aa38eba6f46afddab3a

PR #111, opened 2018-12-14
Fix incorrect link format in data.md

peter50216:
Add a missing []


Gankro:
Fixed in another PR, but thanks!

PR #113, opened 2018-12-16
dropck: The drop order is now defined

vorner:
So the examples are out of date, they no longer pose a problem. However,
there's still a case where it does, so let's use that instead.

Also, mention the definition of drop order in the text, as it is
related.

### DIFF
--- a/src/dropck.md
+++ b/src/dropck.md
@@ -22,25 +22,23 @@ let y;
 }
 ```
 
-Each creates its own scope, clearly establishing that one drops before the
-other. However, what if we do the following?
+There are some more complex situations which are not possible to desugar using
+scopes, but the order is still defined ‒ variables are dropped in the reverse
+order of their definition, fields of structs and tuples in order of their
+definition. There are some more details about order of drop in [rfc1875].
+
+Let's do this:
 
 ```rust,ignore
-let (x, y) = (vec![], vec![]);
+let tuple = (vec![], vec![]);
 ```
 
-Does either value strictly outlive the other? The answer is in fact *no*,
-neither value strictly outlives the other. Of course, one of x or y will be
-dropped before the other, but the actual order is not specified. Tuples aren't
-special in this regard; composite structures just don't guarantee their
-destruction order as of Rust 1.0.
-
-We *could* specify this for the fields of built-in composites like tuples and
-structs. However, what about something like Vec? Vec has to manually drop its
-elements via pure-library code. In general, anything that implements Drop has
-a chance to fiddle with its innards during its final death knell. Therefore
-the compiler can't sufficiently reason about the actual destruction order
-of the contents of any type that implements Drop.
+The left vector is dropped first. But does it mean the right one strictly
+outlives it in the eyes of the borrow checker? The answer to this question is
+*no*. The borrow checker could track fields of tuples separately, but it would
+still be unable to decide what outlives what in case of vector elements, which
+are dropped manually via pure-library code the borrow checker doesn't
+understand.
 
 So why do we care? We care because if the type system isn't careful, it could
 accidentally make dangling pointers. Consider the following simple program:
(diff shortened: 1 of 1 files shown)


src/dropck.md
```diff
@@ -22,79 +22,108 @@ let y;
...
+    fn drop(&mut self) {
+        self.0.borrow_mut().push(self.1);
+    }
+}
+
+let order = RefCell::new(vec![]);
+{
+    let tuple = (PushOrder(&order, 2), PushOrder(&order, 1));
+}
+assert_eq!(order.into_inner(), vec![2, 1]);
```
Askaholic: IMO it's more clear if you switch the values in the tuple so you can make the assertion be:
```rust
assert_eq!(order.into_inner(), vec![1, 2]);
```
That way as you're reading, you know the first item that gets dropped is the one with the value `1`

vorner: I wanted to make it clearer it is in the opposite direction than above, but you're right, it looks rather silly if both are in the reversed order.

Gankro:
Could you explain the motivation for rewriting all the examples? As far as I know nothing has been done that changes their correctness, and I don't think your changes at all make things more clear.

vorner:
It's been a while, so I have to guess a bit. However, the problem with the examples I had is, they still don't compile as they are supposed not to compile. But the reason they do not compile is different and the reason stated on the page is outdated. It's not because one does not outlive the other, but because the order of drop is reversed. If you switch the order in the tuple/pattern (eg. use `(days, inspector)` instead of `(inspector, days)`) it now starts compiling. So the example is not relevant to the chapter at all any more.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=117cf19b8585f2c86f6f530b41a72fc9

I tried to modify the inspector example, but didn't come up with a way. So I created new ones that demonstrate the issue. If you have an idea how to modify the example in a way that doesn't compile in either order of the tuple, I'm all for it.

Gankro:
Could you not just change the examples to `let tup = (a, b)` and using `tup.0` and `tup.1` (or perhaps introduce a struct just for better naming?)

vorner:
> Could you not just change the examples to let tup = (a, b) and using tup.0 and tup.1 (or perhaps introduce a struct just for better naming?)

OK, I'll try playing with it.

vorner:
Just note for myself when I get to it again: the tuple seems to do the trick ‒ this compiles in both orders of the tuple: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7573478e37c067e48a1e728c499773ea

I'll try to return the examples to (close to) the original once I have a bit more time. Thanks for the suggestion.

Gankro:
Just for clarity (named fields) the struct would probably be better, but ok, great!

vorner:
I've returned to to much closer state to the original. It's now wrapped in a structure, to make the borrow checker not being able to track the lifetimes as separate and added references to the defined drop order of fields.

I've force-pushed into the branch, as the diff is shorter to master than to the previous version.

src/dropck.md
```diff
@@ -70,30 +75,38 @@ impl<'a> Drop for Inspector<'a> {
...
     // Let's say `days` happens to get dropped first.
     // Then when Inspector is dropped, it will try to read free'd memory!
 }
 ```
 
 ```text
-error[E0597]: `days` does not live long enough
-  --> src/main.rs:12:28
+error[E0597]: `world.days` does not live long enough
+  --> /tmp/.private/vorner/mdbook-Br8SZ0/dropck.md:88:39
```
Gankro: could you revert this line (for cleanliness)

vorner: Yes, sure (I assume you mean only the one with the file, I'm leaving the addition of `world` in)

Gankro:
thank you!!!!

PR #114, opened 2018-12-26
Fix link in data.md

cmyr:
Just something I noticed reading this morning.


Gankro:
Ah, another one fixed in another PR. Thanks anyway! :D

PR #118, opened 2018-12-29
Trivial updates to the coercions chapter

vorner:
* The modern way is `dyn Trait
* Use `compile_fail` instead of ignore, to make sure it doesn't start
  compiling by accident in some future version of Rust (and also to
  provide link to playground).


matthewjasper:
Thanks!

PR #123, opened 2019-03-08
Fix allocator and pointer issues in vector chapter

cnkei:


src/vec-alloc.md
```diff
@@ -157,7 +155,11 @@ such we will guard against this case explicitly.
 Ok with all the nonsense out of the way, let's actually allocate some memory:
 
 ```rust,ignore
-use std::alloc::oom;
+#![feature(allocator_api, ptr_internals)]
+
+use std::alloc::{Alloc, Global, Layout, handle_alloc_error};
+use std::mem;
+use std::ptr::{NonNull, Unique};
```
Gankro: We should just advise people use NonNull; Unique is used by the stdlib internally but otherwise probably won't ever make it to stabilization.

src/vec-alloc.md
```diff
@@ -157,7 +155,11 @@ such we will guard against this case explicitly.
 Ok with all the nonsense out of the way, let's actually allocate some memory:
 
 ```rust,ignore
-use std::alloc::oom;
+#![feature(allocator_api, ptr_internals)]
```
Gankro: I'm pretty sure you want to just use the stable free functions in std::alloc, which should avoid the need for any features?

https://doc.rust-lang.org/std/alloc/index.html

SimonSapin:
+1 to using `NonNull` and the free functions in `std::alloc`, over `Unique` or `Global`, for code outside of the standard library.

I don’t know if these days it is useful or necessary to keep a `PhantomData<T>` field next to `NonNull<T>`, for variance or dropck or something else. (This is what `Unique<T>` does.)

src/vec-alloc.md
cnkei: I may have overlooked. Will use stable API instead.

cnkei:
Me neither. My knowledge of PhantomData comes from the nomicon :) I will be glad to learn more on this subject.

Gankro:
I'm closing this due to inactivity, but hopefully will get around to this in the near future.

PR #127, opened 2019-03-29
Fix link to type layout reference

GrantMoyer:
I've tested this change with a local static webserver to make sure the link points to the correct place.


GrantMoyer:
Never mind, even though the link is broken [in the actual rust docs][rd], it builds fine here, even in the commit referenced by the rust-lang/rust submodule, f1ff93b.

[rd]: https://doc.rust-lang.org/nomicon/data.html

Gankro:
The docs ride the trains: https://doc.rust-lang.org/nightly/nomicon/data.html

PR #130, opened 2019-04-07
Various unchecked-uninit improvements

vorner:
* Links to the std for some functions
* Mention the MaybeUninit
* Unify indentation (spaces, not tabs)

I was thinking about providing an example with MaybeUninit. But I'm not sure how much it is usable now, reading through the issues it seems things like `[MaybeUninit::uninit(), 50]` doesn't yet work.

Btw, I feel a bit uneasy about typing the links to the std functions, `mdbook serve` doesn't include the std for me so I can't check for them leading anywhere. Is anything checking them in CI?

### DIFF
--- a/src/unchecked-uninit.md
+++ b/src/unchecked-uninit.md
@@ -8,10 +8,10 @@ Unfortunately this is pretty rigid, especially if you need to initialize your
 array in a more incremental or dynamic way.
 
 Unsafe Rust gives us a powerful tool to handle this problem:
-`mem::uninitialized`. This function pretends to return a value when really
-it does nothing at all. Using it, we can convince Rust that we have initialized
-a variable, allowing us to do trickier things with conditional and incremental
-initialization.
+[`mem::uninitialized`][uninitialized]. This function pretends to return a value
+when really it does nothing at all. Using it, we can convince Rust that we have
+initialized a variable, allowing us to do trickier things with conditional and
+incremental initialization.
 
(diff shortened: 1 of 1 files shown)


src/unchecked-uninit.md
```diff
@@ -80,6 +80,19 @@ Every control path through that variable's scope must initialize the value
 before it ends, if it has a destructor.
 *[This includes code panicking](unwinding.html)*.
 
+Not being careful about uninitialized memory often leads to bugs and it has been
+decided the [`mem::uninitialized`][uninitialized] function should be deprecated.
+The [`MaybeUninit`] is supposed to replace it as its API wraps many common
```
Gankro:
```suggestion
The [`MaybeUninit`] type is supposed to replace it as its API wraps many common
```

vorner:
> I think this language needs to be stronger, but this is clearly an improvement, so worth landing.

Do you mean of the newly added paragraph, or of the whole chapter?

Gankro:
my understanding is that mem::uninitialized is borderline impossible to use correctly, because we currently pass metadata to llvm on types with invalid states, like `&T`, and `bool`, which basically instantly trigger UB if a value with that type is `mem::uninitialized()` (llvm undef).

vorner:
Do you have an idea when we can expect `MaybeUninit` to get stabilized? If soon enough, it would probably make sense to rewrite the chapter with that one instead of tuning the warnings about `mem::unitialized`.

Issue #132, opened 2019-04-20
Illegal cpu instruction on OOM?


mexus:
On the [`Allocating Memory`](https://github.com/rust-lang-nursery/nomicon/blob/7a93c089a875bb4287900900c453d7b5133f6e2d/src/vec-alloc.md) page it is said that the `oom` langitem aborts the program by executing an "illegal cpu instruction":

> The standard library calls std::alloc::oom(), which in turn calls the the `oom` langitem. By default this just aborts the program by executing an illegal cpu instruction.

I've checked the code and found out that the `alloc` lib actually calls `std::sys::abort_internal` in [`rust_oom`](https://github.com/rust-lang/rust/blob/master/src/libstd/alloc.rs#L212) (`alloc::alloc::handle_alloc_error` → `std::alloc::rust_oom`), which resolves for example into a `libc::abort` call on unix, which doesn't call any illegal instructions in my understanding but rather sends a `SIGABRT` to the process.

Please correct me if I'm wrong, it might be that I've got the whole idea wrong, but I believe that phrase 

> . By default this just aborts the program by executing an illegal cpu instruction.

should be replaced with something like

> , which aborts the program in a platform-specific manner.

Thanks!

Gankro:
yeah that's probably fair

mexus:
Closing since the PR was merged, thanks!

PR #133, opened 2019-04-20
OOM behaviour in `vec-alloc.md`

mexus:
As per #132


mexus:
It was the fastest merged PR in my entire life. Thanks!

PR #136, opened 2019-05-26
Fix "probably" typo

### LINKED ISSUE #137: "Probably" typo
**Expected (correction bolded):**

> As an aside, while `Send` and `Sync` are unsafe traits, they are also automatically implemented for types when such derivations are pro**b**ably safe to do.

**Actual (typo bolded):**

> As an aside, while `Send` and `Sync` are unsafe traits, they are also automatically implemented for types when such derivations are pro**v**ably safe to do.

Fixed by #136

kylejlin:
Fixes #137


Gankro:
thanks for reporting/fixing this, but provably is indeed the right word here

PR #143, opened 2019-06-19
Fix some links.

ehuss:
rust-lang/rust does not allow bare directory links.
This unblocks updating the submodule.


Gankro:
I don’t understand, those links seem to resolve fine in my browser..?

ehuss:
It's for when viewing offline directly off the filesystem. The browser shows a directory listing instead of loading index.html.

PR #144, opened 2019-06-30
Small README.md cleanup

Lokathor:
* Re-wrap the text to fit in 80 columns.
* Suggest that people use `mdbook build --open` (similar to `cargo doc --open`) to open the book instead of having them worry about what specific OS and browser they're using.
* Make the link to the `mdbook` repo go directly to the current home of `mdbook` in the `rust-lang-nursery`  instead of relying on GitHub to redirect.


Gankro:
thanks!

PR #149, opened 2019-07-20
clarify what is UB

RalfJung:
Also add some missing cases

### DIFF
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -28,6 +28,7 @@ language cares about is preventing the following things:
     * an undefined `enum` discriminant
     * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
     * a non-utf8 `str`
+    * a `NonNull` or `NonZero*` that is 0
     * a compound type (`enum`/`struct`/array/tuple) with an invalid field
 * Unwinding into another language
 * Causing a [data race][race]
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -40,7 +40,7 @@ language cares about is preventing the following things:
       `NonZero*` that is 0
 * Unwinding into another language
 * Causing a [data race][race]
-* Executing code compiled with platform features that the current platform does
+* Executing code compiled with target features that the current thread of execution does
   not support (see [`target_feature`])
 
 "Producing" a value happens any time a value is assigned, passed to a
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -46,10 +46,10 @@ language cares about is preventing the following things:
 "Producing" a value happens any time a value is assigned, passed to a
 function/primitive operation or returned from a function/primitive operation.
 
-A reference/pointer is "dangling" if not all of the bytes it points to are part
-of the same allocation. In particular, null pointers are dangling. The span of bytes it
-points to is determined by the pointer value and the size of the pointee type.
-If the span is empty, "dangling" is the same as "non-null".
+A reference/pointer is "dangling" if it is null or not all of the bytes it
+points to are part of the same allocation. The span of bytes it points to is
+determined by the pointer value and the size of the pointee type.  If the span
+is empty, "dangling" is the same as "non-null".
 
 That's it. That's all the causes of Undefined Behavior baked into Rust. Of
 course, unsafe functions and traits are free to declare arbitrary other
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -23,25 +23,25 @@ language cares about is preventing the following things:
 * Causing a [data race][race]
 * Executing code compiled with [target features][] that the current thread of execution does
   not support
-* Producing invalid primitive values (either alone or as a field of a compound
-  type such as `enum`/`struct`/array/tuple):
+* Producing invalid values (either alone or as a field of a compound type such
+  as `enum`/`struct`/array/tuple):
     * a `bool` that isn't 0 or 1
-    * an undefined `enum` discriminant
-    * null `fn` pointers
+    * an `enum` with an invalid discriminant
+    * a null `fn` pointer
     * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
     * a `!` (all values are invalid for this type)
-    * dangling/unaligned references, references that do themselves point to
-      invalid values, or wide references (to a dynamically sized type) with
-      invalid metadata
+    * a reference that is dangling, unaligned, points to an invalid value, or
+      that has invalid metadata (if wide)
         * slice metadata is invalid if the slice has a total size larger than
           `isize::MAX` bytes in memory
         * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
           `Trait` that matches the actual dynamic trait the reference points to
-    * a non-utf8 `str`
+    * a `str` that isn't valid UTF-8
     * an integer (`i*`/`u*`), floating point value (`f*`), or raw pointer read from
       [uninitialized memory][]
-    * an invalid library type with custom invalid values, such as a `NonNull` or
-      the `NonZero` family of types, that is 0
+    * a type with custom invalid values that is one of those values, such as a
+      `NonNull` that is null. (Requesting custom invalid values is an unstable
+      feature, but some stable libstd types, like `NonNull`, make use of it.)
 
 "Producing" a value happens any time a value is assigned, passed to a
 function/primitive operation or returned from a function/primitive operation.
@@ -50,7 +50,9 @@ A reference/pointer is "dangling" if it is null or not all of the bytes it
 points to are part of the same allocation (so in particular they all have to be
 part of *some* allocation). The span of bytes it points to is determined by the
 pointer value and the size of the pointee type. As a consequence, if the span is
-empty, "dangling" is the same as "non-null".
+empty, "dangling" is the same as "non-null". Note that slices point to their
+entire range, so it's very important that the length metadata is never too
+large. If for some reason this is too cumbersome, consider using raw pointers.
 
 That's it. That's all the causes of Undefined Behavior baked into Rust. Of
 course, unsafe functions and traits are free to declare arbitrary other
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -21,8 +21,8 @@ language cares about is preventing the following things:
 * Breaking the [pointer aliasing rules][]
 * Unwinding into another language
 * Causing a [data race][race]
-* Executing code compiled with target features that the current thread of execution does
-  not support (see [`target_feature`][])
+* Executing code compiled with [target features][] that the current thread of execution does
+  not support
 * Producing invalid primitive values (either alone or as a field of a compound
   type such as `enum`/`struct`/array/tuple):
     * a `bool` that isn't 0 or 1
@@ -81,4 +81,4 @@ these problems are considered impractical to categorically prevent.
 [pointer aliasing rules]: references.html
 [uninitialized memory]: uninitialized.html
 [race]: races.html
-[`target_feature`]: ../reference/attributes/codegen.html#the-target_feature-attribute
+[target features]: ../reference/attributes/codegen.html#the-target_feature-attribute
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -17,7 +17,7 @@ Unlike C, Undefined Behavior is pretty limited in scope in Rust. All the core
 language cares about is preventing the following things:
 
 * Dereferencing (using the `*` operator on) null, dangling, or unaligned
-  pointers
+  pointers, or fat pointers with invalid metadata (see below)
 * Reading [uninitialized memory][]
 * Breaking the [pointer aliasing rules][]
 * Producing invalid primitive values (either alone or as a field of a compound
@@ -30,6 +30,10 @@ language cares about is preventing the following things:
     * dangling/null/unaligned references, references that do themselves point to
       invalid values, or fat references (to a dynamically sized type) with
       invalid metadata
+        * slice metadata is invalid if the slice has a total size larger than
+          `isize::MAX` bytes in memory
+        * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
+          `Trait` that matches the actual dynamic trait the reference points to
     * a non-utf8 `str`
     * an uninitialized integer (`i*`/`u*`) or floating point value (`f*`)
     * an invalid library type with custom invalid values, such as a `NonNull` or
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -21,7 +21,7 @@ language cares about is preventing the following things:
   `enum`/`struct`/array/slice/tuple field address
 * Reading [uninitialized memory][]
 * Breaking the [pointer aliasing rules][]
-* Producing/obtaining invalid primitive values:
+* Producing invalid primitive values:
     * dangling/null/unaligned references
     * null `fn` pointers
     * a `bool` that isn't 0 or 1
@@ -32,6 +32,9 @@ language cares about is preventing the following things:
     * a compound type (`enum`/`struct`/array/tuple) with an invalid field
 * Unwinding into another language
 * Causing a [data race][race]
+
+"Producing" a value happens any time a value is assigned, passed to a
+function/primitive operation or returned from a function/primitive operation.
 
 That's it. That's all the causes of Undefined Behavior baked into Rust. Of
 course, unsafe functions and traits are free to declare arbitrary other
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -21,7 +21,8 @@ language cares about is preventing the following things:
   `enum`/`struct`/array/slice/tuple field address
 * Reading [uninitialized memory][]
 * Breaking the [pointer aliasing rules][]
-* Producing invalid primitive values:
+* Producing invalid primitive values (either alone or as a field of a compound
+  type such as `enum`/`struct`/array/tuple):
     * dangling/null/unaligned references
     * null `fn` pointers
     * a `bool` that isn't 0 or 1
@@ -29,7 +30,6 @@ language cares about is preventing the following things:
     * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
     * a non-utf8 `str`
     * a `NonNull` or `NonZero*` that is 0
-    * a compound type (`enum`/`struct`/array/tuple) with an invalid field
 * Unwinding into another language
 * Causing a [data race][race]
 
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -35,7 +35,8 @@ language cares about is preventing the following things:
         * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
           `Trait` that matches the actual dynamic trait the reference points to
     * a non-utf8 `str`
-    * an uninitialized integer (`i*`/`u*`) or floating point value (`f*`)
+    * an uninitialized integer (`i*`/`u*`), floating point value (`f*`), or raw
+      pointer
     * an invalid library type with custom invalid values, such as a `NonNull` or
       `NonZero*` that is 0
 * Unwinding into another language
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -22,13 +22,14 @@ language cares about is preventing the following things:
 * Breaking the [pointer aliasing rules][]
 * Producing invalid primitive values (either alone or as a field of a compound
   type such as `enum`/`struct`/array/tuple):
+    * a `bool` that isn't 0 or 1
+    * an undefined `enum` discriminant
+    * null `fn` pointers
+    * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
+    * a `!` (all values are invalid for this type)
     * dangling/null/unaligned references, references that do themselves point to
       invalid values, or fat references (to a dynamically sized type) with
       invalid metadata
-    * null `fn` pointers
-    * a `bool` that isn't 0 or 1
-    * an undefined `enum` discriminant
-    * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
     * a non-utf8 `str`
     * an uninitialized integer (`i*`/`u*`) or floating point value (`f*`)
     * an invalid library type with custom invalid values, such as a `NonNull` or
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -16,7 +16,8 @@ to your program. You definitely *should not* invoke Undefined Behavior.
 Unlike C, Undefined Behavior is pretty limited in scope in Rust. All the core
 language cares about is preventing the following things:
 
-* Dereferencing null, dangling, or unaligned references or raw pointers
+* Loading from or storing to null, dangling, or unaligned references or raw
+  pointers
 * Performing out-of-bounds arithmetic for the computation of an
   `enum`/`struct`/array/slice/tuple field address
 * Reading [uninitialized memory][]
@@ -36,6 +37,10 @@ language cares about is preventing the following things:
 
 "Producing" a value happens any time a value is assigned, passed to a
 function/primitive operation or returned from a function/primitive operation.
+
+A reference/pointer is "dangling" if not all of the bytes it points to are part
+of the same allocation. The span of bytes it points to is determined by the
+pointer value and the size of the pointee type.
 
 That's it. That's all the causes of Undefined Behavior baked into Rust. Of
 course, unsafe functions and traits are free to declare arbitrary other
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -22,16 +22,21 @@ language cares about is preventing the following things:
 * Breaking the [pointer aliasing rules][]
 * Producing invalid primitive values (either alone or as a field of a compound
   type such as `enum`/`struct`/array/tuple):
-    * dangling/null/unaligned references
+    * dangling/null/unaligned references, references that do themselves point to
+      invalid values, or fat references (to a dynamically sized type) with
+      invalid metadata
     * null `fn` pointers
     * a `bool` that isn't 0 or 1
     * an undefined `enum` discriminant
     * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
     * a non-utf8 `str`
+    * an uninitialized integer (`i*`/`u*`) or floating point value (`f*`)
     * an invalid library type with custom invalid values, such as a `NonNull` or
       `NonZero*` that is 0
 * Unwinding into another language
 * Causing a [data race][race]
+* Executing code compiled with platform features that the current platform does
+  not support (see [`target_feature`])
 
 "Producing" a value happens any time a value is assigned, passed to a
 function/primitive operation or returned from a function/primitive operation.
@@ -69,3 +74,4 @@ these problems are considered impractical to categorically prevent.
 [pointer aliasing rules]: references.html
 [uninitialized memory]: uninitialized.html
 [race]: races.html
+[`target_feature`]: ../reference/attributes/codegen.html#the-target_feature-attribute
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -16,8 +16,8 @@ to your program. You definitely *should not* invoke Undefined Behavior.
 Unlike C, Undefined Behavior is pretty limited in scope in Rust. All the core
 language cares about is preventing the following things:
 
-* Dereferencing (using the `*` operator on) null, dangling, or unaligned
-  pointers, or wide pointers with invalid metadata (see below)
+* Dereferencing (using the `*` operator on) dangling, or unaligned pointers, or
+  wide pointers with invalid metadata (see below)
 * Reading [uninitialized memory][]
 * Breaking the [pointer aliasing rules][]
 * Unwinding into another language
@@ -31,7 +31,7 @@ language cares about is preventing the following things:
     * null `fn` pointers
     * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
     * a `!` (all values are invalid for this type)
-    * dangling/null/unaligned references, references that do themselves point to
+    * dangling/unaligned references, references that do themselves point to
       invalid values, or wide references (to a dynamically sized type) with
       invalid metadata
         * slice metadata is invalid if the slice has a total size larger than
@@ -48,8 +48,9 @@ language cares about is preventing the following things:
 function/primitive operation or returned from a function/primitive operation.
 
 A reference/pointer is "dangling" if not all of the bytes it points to are part
-of the same allocation. The span of bytes it points to is determined by the
-pointer value and the size of the pointee type.
+of the same allocation. In particular, null pointers are dangling. The span of bytes it
+points to is determined by the pointer value and the size of the pointee type.
+If the span is empty, "dangling" is the same as "non-null".
 
 That's it. That's all the causes of Undefined Behavior baked into Rust. Of
 course, unsafe functions and traits are free to declare arbitrary other
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -17,7 +17,7 @@ Unlike C, Undefined Behavior is pretty limited in scope in Rust. All the core
 language cares about is preventing the following things:
 
 * Dereferencing (using the `*` operator on) null, dangling, or unaligned
-  references or raw pointers
+  pointers
 * Reading [uninitialized memory][]
 * Breaking the [pointer aliasing rules][]
 * Producing invalid primitive values (either alone or as a field of a compound
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -16,16 +16,19 @@ to your program. You definitely *should not* invoke Undefined Behavior.
 Unlike C, Undefined Behavior is pretty limited in scope in Rust. All the core
 language cares about is preventing the following things:
 
-* Dereferencing null, dangling, or unaligned pointers
+* Dereferencing null, dangling, or unaligned references or raw pointers
+* Performing out-of-bounds arithmetic for the computation of a struct/tuple
+  field address
 * Reading [uninitialized memory][]
 * Breaking the [pointer aliasing rules][]
-* Producing invalid primitive values:
-    * dangling/null references
+* Producing/obtaining invalid primitive values:
+    * dangling/null/unaligned references
     * null `fn` pointers
     * a `bool` that isn't 0 or 1
     * an undefined `enum` discriminant
     * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
-    * A non-utf8 `str`
+    * a non-utf8 `str`
+    * a compound type (`enum`/`struct`/array/tuple) with an invalid field
 * Unwinding into another language
 * Causing a [data race][race]
 
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -29,7 +29,8 @@ language cares about is preventing the following things:
     * an undefined `enum` discriminant
     * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
     * a non-utf8 `str`
-    * a `NonNull` or `NonZero*` that is 0
+    * an invalid library type with custom invalid values, such as a `NonNull` or
+      `NonZero*` that is 0
 * Unwinding into another language
 * Causing a [data race][race]
 
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -47,9 +47,10 @@ language cares about is preventing the following things:
 function/primitive operation or returned from a function/primitive operation.
 
 A reference/pointer is "dangling" if it is null or not all of the bytes it
-points to are part of the same allocation. The span of bytes it points to is
-determined by the pointer value and the size of the pointee type.  If the span
-is empty, "dangling" is the same as "non-null".
+points to are part of the same allocation (so in particular they all have to be
+part of *some* allocation). The span of bytes it points to is determined by the
+pointer value and the size of the pointee type. As a consequence, if the span is
+empty, "dangling" is the same as "non-null".
 
 That's it. That's all the causes of Undefined Behavior baked into Rust. Of
 course, unsafe functions and traits are free to declare arbitrary other
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -17,8 +17,8 @@ Unlike C, Undefined Behavior is pretty limited in scope in Rust. All the core
 language cares about is preventing the following things:
 
 * Dereferencing null, dangling, or unaligned references or raw pointers
-* Performing out-of-bounds arithmetic for the computation of a struct/tuple
-  field address
+* Performing out-of-bounds arithmetic for the computation of an
+  `enum`/`struct`/array/slice/tuple field address
 * Reading [uninitialized memory][]
 * Breaking the [pointer aliasing rules][]
 * Producing/obtaining invalid primitive values:
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -18,12 +18,11 @@ language cares about is preventing the following things:
 
 * Dereferencing (using the `*` operator on) dangling, or unaligned pointers, or
   wide pointers with invalid metadata (see below)
-* Reading [uninitialized memory][]
 * Breaking the [pointer aliasing rules][]
 * Unwinding into another language
 * Causing a [data race][race]
 * Executing code compiled with target features that the current thread of execution does
-  not support (see [`target_feature`])
+  not support (see [`target_feature`][])
 * Producing invalid primitive values (either alone or as a field of a compound
   type such as `enum`/`struct`/array/tuple):
     * a `bool` that isn't 0 or 1
@@ -39,8 +38,8 @@ language cares about is preventing the following things:
         * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
           `Trait` that matches the actual dynamic trait the reference points to
     * a non-utf8 `str`
-    * an uninitialized integer (`i*`/`u*`), floating point value (`f*`), or raw
-      pointer
+    * an integer (`i*`/`u*`), floating point value (`f*`), or raw pointer read from
+      [uninitialized memory][]
     * an invalid library type with custom invalid values, such as a `NonNull` or
       the `NonZero` family of types, that is 0
 
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -16,10 +16,8 @@ to your program. You definitely *should not* invoke Undefined Behavior.
 Unlike C, Undefined Behavior is pretty limited in scope in Rust. All the core
 language cares about is preventing the following things:
 
-* Loading from or storing to null, dangling, or unaligned references or raw
-  pointers
-* Performing out-of-bounds arithmetic for the computation of an
-  `enum`/`struct`/array/slice/tuple field address
+* Dereferencing (using the `*` operator on) null, dangling, or unaligned
+  references or raw pointers
 * Reading [uninitialized memory][]
 * Breaking the [pointer aliasing rules][]
 * Producing invalid primitive values (either alone or as a field of a compound
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -17,9 +17,13 @@ Unlike C, Undefined Behavior is pretty limited in scope in Rust. All the core
 language cares about is preventing the following things:
 
 * Dereferencing (using the `*` operator on) null, dangling, or unaligned
-  pointers, or fat pointers with invalid metadata (see below)
+  pointers, or wide pointers with invalid metadata (see below)
 * Reading [uninitialized memory][]
 * Breaking the [pointer aliasing rules][]
+* Unwinding into another language
+* Causing a [data race][race]
+* Executing code compiled with target features that the current thread of execution does
+  not support (see [`target_feature`])
 * Producing invalid primitive values (either alone or as a field of a compound
   type such as `enum`/`struct`/array/tuple):
     * a `bool` that isn't 0 or 1
@@ -28,7 +32,7 @@ language cares about is preventing the following things:
     * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
     * a `!` (all values are invalid for this type)
     * dangling/null/unaligned references, references that do themselves point to
-      invalid values, or fat references (to a dynamically sized type) with
+      invalid values, or wide references (to a dynamically sized type) with
       invalid metadata
         * slice metadata is invalid if the slice has a total size larger than
           `isize::MAX` bytes in memory
@@ -38,11 +42,7 @@ language cares about is preventing the following things:
     * an uninitialized integer (`i*`/`u*`), floating point value (`f*`), or raw
       pointer
     * an invalid library type with custom invalid values, such as a `NonNull` or
-      `NonZero*` that is 0
-* Unwinding into another language
-* Causing a [data race][race]
-* Executing code compiled with target features that the current thread of execution does
-  not support (see [`target_feature`])
+      the `NonZero` family of types, that is 0
 
 "Producing" a value happens any time a value is assigned, passed to a
 function/primitive operation or returned from a function/primitive operation.


src/what-unsafe-does.md
```diff
@@ -16,16 +16,19 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
-    * dangling/null references
+* Producing/obtaining invalid primitive values:
+    * dangling/null/unaligned references
     * null `fn` pointers
     * a `bool` that isn't 0 or 1
     * an undefined `enum` discriminant
     * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
-    * A non-utf8 `str`
+    * a non-utf8 `str`
+    * a compound type (`enum`/`struct`/array/tuple) with an invalid field
```
Gankro: this doesn’t seem to add anything?

Lokathor:
I think the `str` requirement isn't quite correct. As I understand it: the `str` must be valid bytes any time a string processing function is called on it, but during your own `unsafe` blocks it can temporarily be invalid the same as a the fields of a Vec can temporarily be invalid as long as you fix it all by the end of the `unsafe` block.

Lokathor:
Separate post since it's a separate issue and much less prone to discussion:

* the NonNull and NonZero types must not be 0

RalfJung:
> I think the str requirement isn't quite correct. As I understand it: the str must be valid bytes any time a string processing function is called on it, but during your own unsafe blocks it can temporarily be invalid the same as a the fields of a Vec can temporarily be invalid as long as you fix it all by the end of the unsafe block.

I agree but that seems like a discussion separate from this PRs.

> the NonNull and NonZero types must not be 0

Good points. We have some more such types inside rustc and nightly code can define their own if they enable `feature(rustc_attributes)`.

src/what-unsafe-does.md
```diff
@@ -16,16 +16,19 @@ to your program. You definitely *should not* invoke Undefined Behavior.
 Unlike C, Undefined Behavior is pretty limited in scope in Rust. All the core
 language cares about is preventing the following things:
 
-* Dereferencing null, dangling, or unaligned pointers
+* Dereferencing null, dangling, or unaligned references or raw pointers
+* Performing out-of-bounds arithmetic for the computation of a struct/tuple
+  field address
```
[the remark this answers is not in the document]
RalfJung: I don't know what you are asking. How is that UB, or how does one do the arithmetic?

src/what-unsafe-does.md
```diff
@@ -16,16 +16,19 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
 * Reading [uninitialized memory][]
 * Breaking the [pointer aliasing rules][]
-* Producing invalid primitive values:
-    * dangling/null references
+* Producing/obtaining invalid primitive values:
```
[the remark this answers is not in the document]
RalfJung: The key is that
```rust
let b: bool = mem::uninitialized()`
```
or
```rust
let b: bool = *(ptr as *const bool)
```
both trigger this clause. Do you think "produce" covers them?

I should probably add a definition of the term somewhere in this file.

src/what-unsafe-does.md
RalfJung: Previously it said "produce an primitive value". But it is also UB to do
```rust
let b: (bool, u8) = mem::uninitialized();
```
even though this does not really produce any primitive value. I think it is important to stress that we proceed recursively in compound data types. I am happy for suggestions for how to word this better.

Lokathor: It produces a `bool` though?

src/what-unsafe-does.md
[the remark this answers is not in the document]
Lokathor: I think producing covers it.

Linguistically, you can't obtain something until someone has given it to you, and they need to have produced it to give it to you.

src/what-unsafe-does.md
RalfJung: I think this is worth stating explicitly. IMO the previous statement could be interpreted as if creating a compound value would never be a problem, as if wrapping a type in a struct would "hide" the inner invariant -- which I think is a position someone might have. So we should be explicit about this.

Lokathor: I agree, but they probably shouldn't go in a list about primitive values. Instead, change the heading of the list:

> Producing invalid primitive values (either alone or as a field of a compound type):
> * foo..
> * bar..

src/what-unsafe-does.md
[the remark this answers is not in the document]
Gankro: Yeah I can't see how to do this in a way that *wouldn't* also apply to safe references but isn't just a call to an unsafe library function (which all define their own UB contracts and aren't in scope for this page).

src/what-unsafe-does.md
Gankro: yep, that seems a lot better

Gankro:
NonNull and NonZero are again library contracts, and not intrinsic to the language, afaict.

src/what-unsafe-does.md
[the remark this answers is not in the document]
Lokathor: That's a good point if I understand your meaning: "references" doesn't belong on line 19 because they already can't be null, dangling, or unaligned.

Lokathor:
So is the `str` contract. Either it has to get out of the list or they have to be in the list.

Lokathor:
I take it back:

* NonZero/NonNull is definitely a Validity Invariant, while "must be utf8" is either a Validity or Safety invariant (probably safety).
* So it's definitely insta-UB to make a zero NonZeroFoo the same as it is to make a bool that's not 0 or 1.
* So they definitely need to be in the list even if `str` is allowed to be taken off the list for being "only" a safety invariant.

Unfortunately the description doesn't currently talk about validity invariant vs safety invariant, which is maybe a bigger change than this PR was setting out to clean up. However, I think that's the best way for this part of the book to move towards.

src/what-unsafe-does.md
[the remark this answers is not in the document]
Gankro: Right yeah, I think this was written before we came to the conclusion that the *existence* of bad references was immediate UB (the golden days of access-based validity....)

Gankro:
re: str

My understanding is that constructing an invalid str is technically insta-UB, just like bool. Although it's very implausible for the compiler to detect this or use it productively. Has that changed? Is the stdlib or something else actively relying on transiently invalid strs? But more importantly, str is a language primitive, and not something the stdlib cooked up.

re: NonZero

That's a good point on them being insta-UB to mess up. I guess in my mind they're still just library wrappers around some deeper concept that we have simply refused to stabilize. I think I would rather appeal to some more abstract concept like "producing an invalid library type that defines custom invalid values, like NonZero"

src/what-unsafe-does.md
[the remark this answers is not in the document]
Gankro: oh no wait fuck ralf added the references part brains pls 😖

Lokathor:
The fate of `str` is being discussed here: https://github.com/rust-lang/unsafe-code-guidelines/issues/78

That said, I think this whole page needs to back up a moment, talk about Safety Invariant vs Validity Invariant, and then talk about all major `core` types in terms of those definitions.

src/what-unsafe-does.md
RalfJung: Ah, I meant to change the header but forgot.

RalfJung:
> That said, I think this whole page needs to back up a moment, talk about Safety Invariant vs Validity Invariant, and then talk about all major core types in terms of those definitions.

To do that we pretty much have to RFC those notions. I think we can do an update that improves the docs without going through all of that.

So, I agree that should happen eventually -- but we are talking about a timescale of months here.

> That's a good point on them being insta-UB to mess up. I guess in my mind they're still just library wrappers around some deeper concept that we have simply refused to stabilize.

In some sense that's true, and the deeper concept currently has the name `rustc_layout_scalar_valid_range_start`. But I did not want to mention *that* here.

src/what-unsafe-does.md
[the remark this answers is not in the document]
RalfJung: > (the golden days of access-based validity....)

You mean when what we documented as UB and what we compiled for LLVM were not matching? ;)

Indeed for references I *think* this is somewhat overlapping with the rule that you cannot even create dangling references, but for raw pointers it is not. I am also not sure about references to unsized types.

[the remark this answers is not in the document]
RalfJung: So do you want me to change something here now?

src/what-unsafe-does.md
```diff
@@ -16,19 +16,32 @@ to your program. You definitely *should not* invoke Undefined Behavior.
 Unlike C, Undefined Behavior is pretty limited in scope in Rust. All the core
 language cares about is preventing the following things:
 
-* Dereferencing null, dangling, or unaligned pointers
+* Loading from or storing to null, dangling, or unaligned references or raw
+  pointers
```
RalfJung: I changed "dereferencing" into "loading from or storing to", because if a dangling/unaligned pointer is dereferenced (as in `*ptr` is syntactically written) without incurring a memory access (as in `&(*ptr).field`), that does not actually cause UB to my knowledge.

src/what-unsafe-does.md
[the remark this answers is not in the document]
Gankro: I would drop mention of references here (reverting to the checked-in state); if you're not allowed to create bad ones, using bad ones is irrelevant.

And I would completely drop any mention of offsets unless you can provide a case where this happens with raw pointers and no other rule covers and it's *not* about using offset/add (they document it, and wrapping_offset specifically allows it, so it is in fact not intrinsic to the language).

Are you picturing something like creating a type whose `size_of` is `> usize::MAX` (does the compiler even permit this??), and relying on a field access to wrap around the address space and alias some earlier part of the type???

src/what-unsafe-does.md
Gankro: That's a pretty subtle claim! I'm not sure the nomicon can justifiably document something so subtle. As a general rule, we just say more things are UB than *actually* are, because it is of course completely fine for devs to be more careful than the really need to be, and it's back-compat to suddenly say "this is now *NOT* UB"

src/what-unsafe-does.md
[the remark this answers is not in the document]
RalfJung: I think you could do something like this:
```rust
fn main() {
    let local = 5u8;
    let ptr = (&local as *const u8).wrapping_sub(1) as *const (u8, u8);
    let _ref = unsafe { &(*ptr).1 };
}
```
This should be UB but I don't think it violates any of the other clauses?

Also with https://github.com/rust-lang/rfcs/pull/2582 it is easy to construct an example:
```rust
fn main() {
    let ptr: *const (u8, u8) = 16usize as *const _;
    let ptr2 = &raw const (*ptr).1;
}
```

src/what-unsafe-does.md
RalfJung: > As a general rule, we just say more things are UB than actually are

The problem is that this can lead to people just ignoring what we say because "it says too much UB anyway".

This is also somewhat related to https://github.com/rust-lang/rfcs/pull/2582.

src/what-unsafe-does.md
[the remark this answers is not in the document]
RalfJung: But I guess this relates to the other discussion -- they might both be UB because a dangling raw ptr gets "dereferenced" even though no access happens. I thought it was established that we wanted to require as little as possible for raw pointers.

Funny how I am arguing for an access-based model now and you are arguing against it... ;)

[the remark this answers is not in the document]
Lokathor: Doesn't the fact that it does wrapping_sub to go out of bounds of the allocated object make it UB to use that pointer?

I mean I know that wrapping_sub allows the math to go out of bounds without being UB, but the resulting pointer is invalid if out of bounds, right?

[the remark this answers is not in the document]
RalfJung: Using `wrapping_add` it is certainly possible to make this a valid pointer again.

src/what-unsafe-does.md
Lokathor: Except that you've also said yourself before, I think it was on Zulip recently: Until something says you _can_ do something, it's UB by default.

RalfJung: That's *my* position, but my experience tell me other people take a more... "relaxed" approach to UB.

My main reason that I don't thing dereferencing a dangling pointer should be UB is that I hope to, some day, clean up [this mess](https://github.com/rust-lang/rust/blob/95b1fe560d2bd8472f250fb8cfd2168520a58405/src/liballoc/collections/btree/node.rs#L594-L609). But to do that I need to be able to take a raw pointer to a large type that is partially dangling, and still compute the address of a field in there that is in the non-dangling part.

src/what-unsafe-does.md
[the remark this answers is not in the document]
Lokathor: Uh, right. "use" is too colloquial. "access through" is the term I suppose.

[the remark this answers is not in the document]
RalfJung: At least in my dictionary, the two examples above however don't "access" through the raw ptr. They do a ptr-to-place conversion (that's the `*` operator), then restrict the place to a field, and then a place-to-ptr coercion (the `&` operator). No memory access.

Lokathor:
So, in that case I think my question becomes... is Rustonomicon itself RFC controlled or informal information? Is the information "how it works today" or is it "stable for all time"?

If it's not RFC controlled stable, just a best effort to document the innards of Rust as it exists today, I don't see the harm in pointing at even draft definitions, assuming that the draft is also its own best effort, if that gives people a much better way to talk about when the UB will happen to them.

The concept of a Validity Invariant and how it's insta-UB to set `x: bool = transmute(2);` even if you never look at it, compared to a Safety Invariant where you can call `Vec::set_len(&core::usize::MAX)` and it's fine _until you use the vec wrong_ is extremely _teachable_ because it gives names to use to talk about things, so students can ask about what they really mean when they have a question.

Either way, we don't need to add all that in this PR right now, but I'd like to know how official and immutable Rustonomicon is trying to be, and where it lives compared to the Reference and the UCG.

src/what-unsafe-does.md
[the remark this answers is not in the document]
Lokathor: hmmm, this is a tricky one. I'll try and sit down at my desktop and do the heavy thinking thing later today.

Gankro:
The Rustonomicon is non-normative and exists primarily to allow people write correct programs. The *most* important thing is that if we say "this is OK to do" then we are in fact correct. It is perfectly fine if we say that you *can't* do something if you in fact really can. In particular, the world cannot wait in baited breath for the UCG team to formalize the entire language (a hard and time-consuming task), so the nomicon conservatively declares things to be UB which might not be.

In some places we have taken shots on calling things Correct that are ultimately declared Wrong or dubious (e.g. access-based validity). Ideally, this is based on Real Things The Stdlib Does. At least in those cases we're *probably* not going to misoptimize you, and hopefully the fact that it's wrong will cause enough of a ruckus in the stdlib for it to become A Well-Known Issue.

I am sympathetic to the desire to be maximally accurate so that devs won't dismiss claims as obviously over-broad, and this is definitely a place to be pedantic! But also it might be good in some places to be over-broad just as a matter of clarity.

Talking about Validity vs Safety (or whatever) does indeed sound within scope of the rustonomicon. I'll need to think about it.

src/what-unsafe-does.md
RalfJung: I think I came to the conclusion that for now the Nomicon should probably stick to "dereference", and I think that also means we don't need to "pointer offset" clause.

Some time after we resolved https://github.com/rust-lang/rfcs/pull/2582 (heh), we should officially determine if its "deref" or "access" that counts, and then adjust whatever else needs adjusting.

RalfJung:
I think this resolves the open discussions.

How to incorporate the validity/safety invariant thing seems to me to be a topic for a separate PR.

Havvy:
Reminder to also copy this PR onto the reference after it's merged here.

RalfJung:
@Havvy Oh, you mean for https://doc.rust-lang.org/stable/reference/behavior-considered-undefined.html ? I *thought* that I had seen another "UB list" somewhere but could not remember where.

Is there any good reason to even have two lists?

Lokathor:
Ideally, the reference will become less generally PR-able and drifts towards being fully normative over time.

At the moment, the reference is just as non-normative as the Nomicon and UCG, so there's not a big deal.

RalfJung:
In fact the nomicon looks more normative because the reference has all these big fat red warnings everywhere.^^

But I was mostly concerned about having to keep the lists in sync.

Lokathor:
I still think that the bullet point

> * Dereferencing (using the `*` operator on) null, dangling, or unaligned references or raw pointers

Needs to remove the word "references". Line 25 already lists null/dangling/unaligned references as UB, so if you have one to de-reference you're already off the edge of the map.

RalfJung:
My thinking was that it is better to cover a case twice than to forget one.

Lokathor:
"It wouldn't say that you can't dereference them if you weren't even allowed to have them, it's fine"
"dude what about the list a few lines down?"
"Look it wouldn't talk about unaligned references if they were impossible, so I'm not gonna change it"

when this conversation inevitably happens I'm gonna link them this PR thread ;3

RalfJung:
When we inevitably remove a UB clause because we think it is unreachable, but in fact it isn't, I'm going to blame you. ;)

RalfJung:
But in the end I could live with either way here, so I'll just wait for what @Gankro thinks.

Gankro:
the "schtick" of this page is "wow there's not much intrinsic UB you need to know about, what a simple language" so I'd prefer a minimal and orthogonal definition 😊

RalfJung:
Okay, done that.

RalfJung:
In terms of making sure we definitely do not call out "too little" UB, there are still some things coming to my mind that we then might want to add, somehow:

* We do not talk about "fat references" and what the requirement on the metadata part of them is. See https://github.com/rust-lang/unsafe-code-guidelines/issues/166.
* `let u8 = mem::uninitialized();` [might be UB](https://github.com/rust-lang/unsafe-code-guidelines/issues/71) if we decide against allowing uninitialized integers.
* References, beyond not dangling, [might actually have to point to valid data](https://github.com/rust-lang/unsafe-code-guidelines/issues/77). In particular there is some desire for making producing an `&!` UB.

src/what-unsafe-does.md
```diff
@@ -16,19 +16,30 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
     * null `fn` pointers
     * a `bool` that isn't 0 or 1
     * an undefined `enum` discriminant
     * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
-    * A non-utf8 `str`
+    * a non-utf8 `str`
+    * an invalid library type with custom invalid values, such as a `NonNull` or
+      `NonZero*` that is 0
 * Unwinding into another language
 * Causing a [data race][race]
```
gnzlbg: Calling a `#[target_feature(enable = "foo")] unsafe fn bar();` function on a thread of execution that does not support the target-feature `foo` is undefined behavior as well. 

It's not clear to me whether the `target_feature` attribute makes this a precondition on calling the function - all these functions are `unsafe fn` anyways - or whether this is something worth documenting here.

RalfJung: Good point; this is listed in the other UB list at https://doc.rust-lang.org/reference/behavior-considered-undefined.html. @Gankro what do you think?

RalfJung:
Okay I think this now covers any UB that primitive operations (as opposed to `unsafe fn`/intrinsics) can cause.

src/what-unsafe-does.md
```diff
@@ -16,18 +16,35 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
+    * dangling/null/unaligned references, references that do themselves point to
+      invalid values, or fat references (to a dynamically sized type) with
+      invalid metadata
+    * a non-utf8 `str`
+    * an uninitialized integer (`i*`/`u*`) or floating point value (`f*`)
+    * an invalid library type with custom invalid values, such as a `NonNull` or
+      `NonZero*` that is 0
 * Unwinding into another language
 * Causing a [data race][race]
+* Executing code compiled with platform features that the current platform does
```
gnzlbg:
```suggestion
* Executing code compiled with target-features that the current thread of execution does
```

These are historically-called target-features but nowadays targets do support different features in different threads of execution, e.g., on heterogeneous SoCs where one core supports features that the others do not.

src/what-unsafe-does.md
```diff
@@ -16,18 +16,40 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
+* Producing invalid primitive values (either alone or as a field of a compound
+  type such as `enum`/`struct`/array/tuple):
     * a `bool` that isn't 0 or 1
     * an undefined `enum` discriminant
+    * null `fn` pointers
     * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
-    * A non-utf8 `str`
+    * a `!` (all values are invalid for this type)
+    * dangling/null/unaligned references, references that do themselves point to
+      invalid values, or fat references (to a dynamically sized type) with
```
Gankro: i think we're trying to phase out fat for wide

```suggestion
      invalid values, or wide references (to a dynamically sized type) with
```

src/what-unsafe-does.md
```diff
@@ -16,18 +16,40 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
+      invalid metadata
+        * slice metadata is invalid if the slice has a total size larger than
+          `isize::MAX` bytes in memory
+        * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
+          `Trait` that matches the actual dynamic trait the reference points to
+    * a non-utf8 `str`
+    * an uninitialized integer (`i*`/`u*`), floating point value (`f*`), or raw
+      pointer
+    * an invalid library type with custom invalid values, such as a `NonNull` or
+      `NonZero*` that is 0
```
Gankro: NonZero* keeps tripping me up because I'm parsing it as a pointer, let's just drop it and only use NonNull as the example.

src/what-unsafe-does.md
```diff
@@ -16,18 +16,40 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
+    * dangling/null/unaligned references, references that do themselves point to
+      invalid values, or fat references (to a dynamically sized type) with
+      invalid metadata
+        * slice metadata is invalid if the slice has a total size larger than
+          `isize::MAX` bytes in memory
+        * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
+          `Trait` that matches the actual dynamic trait the reference points to
+    * a non-utf8 `str`
+    * an uninitialized integer (`i*`/`u*`), floating point value (`f*`), or raw
+      pointer
```
Gankro: I want to think about this precise wording. I *think* you can break this out of here and fold it into "reading uninitialized memory". I want to have some wording along the lines of "you can't tell the compiler a value definitely exists where only uninitialized memory resides", but this needs to be reinforced by documentation.

src/what-unsafe-does.md
```diff
@@ -16,18 +16,40 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
+    * an invalid library type with custom invalid values, such as a `NonNull` or
+      `NonZero*` that is 0
 * Unwinding into another language
 * Causing a [data race][race]
+* Executing code compiled with target features that the current thread of execution does
+  not support (see [`target_feature`])
```
Gankro: Formatting nit: have this above invalid primitives so that the massive sublist is the last thing in the super-list.

src/what-unsafe-does.md
RalfJung: This is a type-based property though. The following is fine:
```rust
let x: MaybeUninit<u8> = mem::uninitialized();
```
but the following is not:
```rust
let x: u8 = mem::uninitialized();
```
I would actually remove the "Reading uninitialized memory" clause in favor of this type-based one.

RalfJung: I gave this a shot, let me know what you think.

RalfJung:
@Gankro any update on this?

src/what-unsafe-does.md
Gankra: so i get what you're saying but two things:

A) Pedantic aside that probably doesn't matter: do we actually care about guaranteeing `let x: MaybeUninit<u8> = mem::uninitialized();` is valid? I suppose that's important if you want to be able to cast a pointer to a fresh allocation to be `&mut MaybeUninit` and not do dummy writes of `uninit()` to be able to call its methods? Hmm...

B) I was using a subtler definition of "a value exists", which agrees with your statement that it's type-based. Basically empty types don't have values; they can't. So by having a `union { (), T }` you are, *with types*, telling the compiler that it may always be the case that a value doesn't exist there (but it's not opaque like asm or volatile, so the compiler can clearly observe reads and writes and reason about the memory's bytes).

Gankra: hmm, my definition of empty things being valueless fails to handle `!`

Gankra:
As discussed in zulip, with the introduction of `&raw` i believe it becomes necessary to explicitly define that it's UB to write to "read only" memory (immutable statics that don't contain UnsafeCell), as the pointer aliasing rules seemingly no longer cover that issue.

As always I would prefer that this case is actually well-defined (as there is a very reasonable safe behaviour on all major platforms with default configurations -- fault and crash). But I am sympathetic to this being potentially hard to properly define/support/guarantee.

Also it's just unclear to me how to define this in a clear way. I'm not 100% comfortable with just dismissing/ignoring the existence of readonly memmap'd pages, as this feels like a *very* important thing to elaborate on.

gnzlbg:
> As discussed in zulip, with the introduction of &raw i believe it becomes necessary to explicitly define that it's UB to write to "read only" memory (immutable statics that don't contain UnsafeCell), as the pointer aliasing rules seemingly no longer cover that issue.

Not only statics:

```rust
let x = 3;
unsafe { *(&raw x as *mut i32) = 42 };
```

does not violate any pointer aliasing rules either AFAICT, yet I feel that 3 can be put in read-only memory.

> As always I would prefer that this case is actually well-defined (as there is a very reasonable safe behaviour on all major platforms with default configurations -- fault and crash). But I am sympathetic to this being potentially hard to properly define/support/guarantee.

Whether faults, crashes, etc. exist depends on the target. The proper thing to do here would be to `abort`, which on some targets would just loop forever. 

I don't know of any target for which this would be cheap, e.g., most targets just SIGSEV on this, and I don't think it would be ok for the Rust runtime to install a signal handler for these and turn those into aborts, in most cases we can't tell where the SIGSEV is coming from. Even if we did that, the user can replace our signal handler, so I don't think we could guarantee an abort even on "nice" targets.

On targets that do not support read-only memory and just put statics into normal memory, we probably would need to check whether each write access the memory of a static at runtime to be able to abort. Optimizations might be able to, in some cases, conclude that a pointer never points to a static, but it feels like a long shot.

We don't even guarantee that a stack overflow aborts the program if it happens in safe Rust. A write to an immutable variable cannot happen in safe Rust, so all of this would feel like too much trouble for something that's inherently unsafe anyways (pointer dereferences). 

This doesn't mean that we shouldn't try to make the program terminate ASAP if a write to an immutable variable happens - undefined behavior allows us to do anything, including emitting a nice error message. I have no idea how we could guarantee that such a things happens without adding overhead.

> Also it's just unclear to me how to define this in a clear way. I'm not 100% comfortable with just dismissing/ignoring the existence of readonly memmap'd pages, as this feels like a very important thing to elaborate on.

The functions of `GlobalAlloc` and `Alloc` state their safety invariant, which tells you what you are allowed to do with the memory they return. 

I don't see how `mmap` would be any different than, e.g., the following function `foo`: 

```c
// You can only read from foo() if the integer 
// representation of the pointer is `42`. 
int8_t* foo();
```

If you do an FFI call, you are on your own. Read the docs, hope that they are correct, and do what they say. Even for mmap, dozens of OSes implement it, all of them with custom extensions. I don't see what could we possibly say about it here, nor how that would be any more useful to users than "read the docs".

src/what-unsafe-does.md
RalfJung: > do we actually care about guaranteeing let x: MaybeUninit<u8> = mem::uninitialized(); is valid?

Yes! It is operationally equivalent to `let x: MaybeUninit<u8> = MaybeUninit::uninit();`, which is safe.

> i'm a bit uncomfortable with technical discussion appealing to mem::uninitialized since it's busted in a bunch of ways, so it doesn't actually have a coherent meaning

I view it as a transmute from a byte-string of appropriate length full of uninitialized bytes, to `T`. That's a well-defined notion on the Abstract Machine. It is equivalent to something like "allocate fresh memory, then call `ptr::read`".

> I was using a subtler definition of "a value exists", which agrees with your statement that it's type-based. 

Could you try to define that? Because I don't really know what you mean.

> Basically empty types don't have values; they can't.

I think thinking specifically about empty types is a red herring. Whatever happens for empty types should fall out of more general rules that also apply to, e.g., `bool`.

RalfJung:
> as the pointer aliasing rules seemingly no longer cover that issue.

I think they can -- and in fact, Stacked Borrows does. But you [may not like the way this happens](https://github.com/rust-lang/rust/issues/56604).

@gnzlbg 
> does not violate any pointer aliasing rules either AFAICT, yet I feel that the behavior of this code is undefined.

I think *if* this is UB it must be through pointer aliasing rules. And, see above -- Stacked Borrows makes that happen.

Gankra:
(we are going to punt on the read-only memory issue for now since &raw doesn't exist yet)

Lokathor:
mutating a non-mut binding because you took a shared borrow on it and then use it as a unique borrow _must_ violate some rule somewhere, but I'm also failing to recall exactly where such a thing is already written.

src/what-unsafe-does.md
Gankra: Here's some Proposed Terminology for the sake of this conversation. This is not necessarily useful for a formal model, but it is helping me make sense of the concepts in my head:

* a **value** is a property a byte or collection of bytes has. Bytes may have the secret 257th  value of "uninitialized". 

* An **instance** of a type is something you get from constructing that type. Moving or destroying an instance relinquishes a piece of code's access to that instance. Those instances do not necessarily have a unique, tracked, identity, but their ability to *reach* a point in the program serves as a kind of proof (more on that below). The Copy trait just gives permission to all code that has acquired an instance to mint more instances with the same value. 

I am not 100% comfortable with the notion of "construction", as room must be made for transmutes, plain-old-data, and deserializers. I think this basically boils down to a similar situation as double-drops where any type is freely allowed to assume its instances aren't double-dropped with UB at stake, but a type may randomly decide it's fine to be double-dropped. There's no trait/property for this, it's just a thing a type can randomly be fine with. It's very contrived to come up with such a type, so we don't care about this much outside of the very specific niche of standards and trivia.

Similarly, types must be able to assume that there is meaning to the existence of an instance, also with UB at stake. But types may allow instances to be minted without any constructor. Which is to say, for some types it may be fine to have more instances of a type "live" in a program than the number of its constructors that have run. This *is* a relatively important fact about a type, but just like with doubledropabble it's not something with a standardized trait/property. (perhaps one can just appeal to access to a type's record initialization syntax as permission to mint as many instances as one pleases without even using that syntax?)

Anyway.

Certain types may have forbidden values, indicating that if a piece of code claims to have an instance of a type, the memory that instance occupies *must not* have certain values.

Zero-sized types do not have values, but they do still have instances. A logical consequence of this is that we cannot have the concept of a zero-sized type with a forbidden value. It doesn't make sense.

From this, I would like to assert that `!` does not in fact fall under the umbrella of forbidden values: it is simply forbidden to claim that you have an instance of `!`, because it does not have a constructor. If you claim to have an instance of `!`, you are lying, and we have made that lie unconditional UB (as is our right as type authors).

Some additional context, from the practical perspective:

In the stdlib we frequently appeal to the idea that, logically speaking, there is always an instance of a zero-sized type everywhere you look. So e.g. `(73 as *const ()).read()` is not UB ("you can pull instances from the aether").

This logically falls out from needing to be able to implement a type like Vec<ZST> which doesn't like, reserve an address in memory for its copies of the ZSTs. It just dumps them into the aether and then pulls them from the aether when it needs to.

However it *is* dangerous to pull values from the aether like this. It is an important property of rust that privacy lets abstraction authors use the *existence* of a instance of a type they control serve as a proof that certain things have occurred, and this extends to ZSTs. So for instance one might use a ZST as a token giving permission to access a global, and so pulling that token from the aether could cause UB.

The obvious rule of thumb here is that you should only be able to pull as many instances from the aether as you dumped there. Hence `Vec<ZST>` is fine, because it is basically just a counter that tracks how many instances of the ZST that this Vec has been given and subsequently dumped into the aether. This counter prevents safe code from pulling extra instances from the aether (forging them).

This jives well with `!` being illegal to claim to have an instance with: you "can't" pull an instance from the aether that you did not dump there (unless you have access to the type's constructor and can therefore mint as many instances as you please).

Gankra: postscript: claiming to have an instance of `union MyUnion { a: T, b: U }` is a claim that you have either an instance of T or an instance of U stored in the instance's memory (this similarly follows for enums, but discriminants make it messier to talk about so let's ignore them). 

If T is a ZST (as in MaybeUninit), it follows that it is always valid for a MyUnion instance to be uninitialized, as that is always conformant with the claim that "there exists an instance of T here". Similarly if T is fine to instantiate, it's fine for U to be illegal to instantiate (`!`), as "p or contradiction" is just "p".

An interesting claim that falls out of what I just wrote here is that if T and U have mutually invalid values, the compiler may conclude that those mutually-invalid values are invalid for a MyUnion to have. So a union (or enum) is not actually completely opaque to the compiler. It's just that if one arm is `()`, the compiler has to let you do whatever you want with the memory/instantiation. I am lightly worried this contradicts something important, but I'm not sure!

RalfJung: > a value is a property a byte or collection of bytes has. Bytes may have the secret 257th value of "uninitialized".

Okay, so that sounds more like what I wold call a "(Rust) Abstract (Machine) Byte".

> An instance of a type is something you get from constructing that type.

Is this a list of bytes, of values, or something more mathematically abstract?

> Zero-sized types do not have values, but they do still have instances. A logical consequence of this is that we cannot have the concept of a ZST with a forbidden value. Nor can we speak of an "uninitialized" ZST. It doesn't make sense.
> 
> From this, I would like to assert that ! does not in fact fall under the umbrella of forbidden values: it is simply forbidden to claim that you have an instance of !, because it does not have a constructor. If you claim to have an instance of !, you are lying, and we have made that lie unconditional UB (as is our right as type authors).

To me this indicates your concepts are flawed. I feel the fact that `!` has 0 "inhabitants" and `()` has 1 and `bool` has 2 should nicely fall out of the same concept. I have [described such a concept in the UCG repo](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/value-domain.md), but it is still work-in-progress.

In particular, "ZSTs have no values" sounds just so wrong to me. The type `()` obviously has 1 value, and that's `()`! (We sadly write them the same in Rust, but you get what I mean.) You even said yourself that "value" has something to do with a "collection of bytes"; if I interpret that freely as "list of bytes" then the type `()` has one "value", and that's the empty list.

"No value" to me sounds like "empty type", and that raises some bad memories about how many people call `void` in C "empty". That's not accurate. `void` is exactly like `()`, it *conveys no information*, but it isn't "empty". The way to formally model "no information" is to say that "there is only one possible answer", AKA "the answer was already known so it doesn't even have to be given", AKA a type with exactly 1 possible value.

Information-theoretically this also works perfectly: you need `log2(N)` many bits to encode which of N answers is given, and `log2(1) = 0`. So you need 0 bits to encode the `()` type.

> In the stdlib we frequently appeal to the idea that, logically speaking, there is always an instance of a zero-sized type everywhere you look. So e.g. (73 as *const ()).read() is not UB ("you can pull instances from the aether").

The way I would phrase it is that copying/moving a ZST is a NOP. There is nothing to copy, operationally. As you said, `Vec` won't pull "more instance of a ZST out of this air as exist in the vector", so logically speaking, this is not *really* out-of-thin air -- it's just that you cannot see the moves that are happening.

We've had [cases where our handling of ZST did not follow this model](https://github.com/rust-lang/rust/pull/52206), but I think those were bugs.

> However it is dangerous to pull values from the aether like this. It is an important property of rust that privacy lets abstraction authors use the existence of a instance of a type they control serve as a proof that certain things have occurred, and this extends to ZSTs. So for instance one might use a ZST as a token giving permission to access a global, and so pulling that token from the aether could cause UB.

Agreed. However, in my terminology, those privacy-induced extra properties arise from the [safety invariant, not the validity invariant](https://rust-lang.github.io/unsafe-code-guidelines/glossary.html#validity-and-safety-invariant).

---

All of that said, I think much of it is not relevant for a discussion about "what is UB". Even if `Vec` *did* "duplicate" a ZST, that would not be instantaneous UB! We cannot make the Abstract Machine understand the safety invariant, it can be arbitrarily complicated after all and is often user-defined. 

In other words, I claim that for any slice of ZST, doubling its length does not cause undefined behavior. Doing this may lead to UB later, when some code that assumed its safety invariant is not broken runs, but the act of doubling the slice length itself is perfectly fine, from the perspective of the Abstract Machine.

---

> claiming to have an instance of union MyUnion { a: T, b: U } is a claim that you have either an instance of T or an instance of U stored in that memory

We have some disagreements here. ;)

I think treating unions this way is very, very complicated because when we go about defining the Abstract Machine in a *concrete, operational way* (like an interpreter), we cannot know which variant the union is in. It is because of that that I am strongly pushing for "unions have no active variant". https://github.com/rust-lang/rfcs/blob/master/text/2514-union-initialization-and-drop.md deliberately did not have such a notion, and these days the [Reference states](https://doc.rust-lang.org/reference/items/unions.html):

*Unions have no notion of an "active field".*

> An interesting claim that falls out of what I just wrote here is that if T and U have mutually invalid values, the compiler may conclude that those mutually-invalid values are invalid for a MyUnion to have. So a union (or enum) is not actually completely opaque to the compiler.

I think you are re-tracing a lot of https://github.com/rust-lang/unsafe-code-guidelines/issues/73 right now. ;) The OP of that thread already contains some code that I (and many in the UCG) think is legal, but that you just ruled out.

Gankra: I didn't properly consider the claim of unions "having an instance" in the context of field-by-field re-initialization into another type (which is of course extremely important). But my intended claim here is exactly the one gnzlbg is championing in that thread (and I have replied there to that effect) -- only if niches align is a niche actually available. Hence "mutually invalid values" (the intersection of their value domains, in your terminology). This is consistent with the way we make you create a union by providing an entire instance of one of the union's cases. Like, by default to create a union you have to very materially have an instance.

But yeah once you start using the union you can enter a super-position of instances, which certainly makes it awkward to talk about. I am *absolutely not* intending to suggest there is an "active member" as that is a horribly cursed concept.

Certainly though, if you ever do `&mut my_union.a`, in my understanding of the current reference semantics, you are making a very material claim of having an instance of that type. And it would also be very difficult for us to make unions *robustly* opaque, in the way asm!() blocks and volatile accesses are.

Something still really bothers about the rules of ! falling out the rules for value domains. Like yes I understand the notion of 0 vs 1 inhabitants (although thank you for reminding me of that term), but inhabitants don't feel like the right abstraction for talking about packing bits? Like you also appealed to the notion of "a list of bytes", which I agree can reasonably include the empty list, but if I ask for a list, you're not allowed to come back with "well, here's no list at all, that's a list right".

Gonna keep reading the relevant things over and mulling it over. I am under the impression I'm just in a weird headspace and I'll eventually come around to your position on just having ! slot in here naturally.

src/what-unsafe-does.md
```diff
@@ -16,18 +16,41 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
+  wide pointers with invalid metadata (see below)
 * Breaking the [pointer aliasing rules][]
-* Producing invalid primitive values:
-    * dangling/null references
-    * null `fn` pointers
+* Unwinding into another language
+* Causing a [data race][race]
+* Executing code compiled with target features that the current thread of execution does
+  not support (see [`target_feature`][])
+* Producing invalid primitive values (either alone or as a field of a compound
```
Gankra: "primitive" should be removed here, as NonNull and friends are not primitives.

src/what-unsafe-does.md
```diff
@@ -16,18 +16,41 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
+      the `NonZero` family of types, that is 0
+
+"Producing" a value happens any time a value is assigned, passed to a
+function/primitive operation or returned from a function/primitive operation.
+
+A reference/pointer is "dangling" if it is null or not all of the bytes it
+points to are part of the same allocation (so in particular they all have to be
+part of *some* allocation). The span of bytes it points to is determined by the
+pointer value and the size of the pointee type. As a consequence, if the span is
+empty, "dangling" is the same as "non-null".
```
Gankra: Extra detail to add at the end.

Note that slices point to their *entire* range, so it's very important that the length metadata is never too large. If for some reason this is too cumbersome, consider using raw pointers.

src/what-unsafe-does.md
```diff
@@ -16,18 +16,41 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
+        * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
+          `Trait` that matches the actual dynamic trait the reference points to
+    * a non-utf8 `str`
+    * an integer (`i*`/`u*`), floating point value (`f*`), or raw pointer read from
+      [uninitialized memory][]
+    * an invalid library type with custom invalid values, such as a `NonNull` or
+      the `NonZero` family of types, that is 0
+
+"Producing" a value happens any time a value is assigned, passed to a
+function/primitive operation or returned from a function/primitive operation.
```
Gankra: suggested reword and massive clarification:

Many have trouble accepting the consequences of invalid values, so they merit some extra discussion. The claim being made here is a very strong one, so read carefully.

A value is *produced* whenever it is assigned, passed to something, or returned from something. Keep in mind references get to assume their referents are valid, so you can't even create a reference to an invalid value. Additionally, uninitialized memory is **always invalid**, so you can't assign it to anything, pass it to anything, return it from anything, or take a reference to it. (Padding bytes are not technically part of a value's memory, and so may be left uninitialized.)

In simple and blunt terms: you cannot ever even **suggest** the existence of an invalid value. No, it's not ok if you "don't use" or "don't read" the value. Invalid values are **instant Undefined Behaviour**. The only correct way to manipulate memory that could be invalid is with raw pointers using methods like `write` and `copy`. If you want to leave a local variable or struct field uninitialized (or otherwise invalid), you *must* use a union or enum which clearly indicates at the type level that this memory may contain no values (see MaybeUninit for details).

src/what-unsafe-does.md
```diff
@@ -16,18 +16,41 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
 
-* Dereferencing null, dangling, or unaligned pointers
-* Reading [uninitialized memory][]
+* Dereferencing (using the `*` operator on) dangling, or unaligned pointers, or
+  wide pointers with invalid metadata (see below)
 * Breaking the [pointer aliasing rules][]
-* Producing invalid primitive values:
-    * dangling/null references
-    * null `fn` pointers
+* Unwinding into another language
```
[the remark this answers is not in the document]
rkruppe: It is currently de facto undefined to unwind out of a Rust function defined as `extern "C"` no matter the caller, because our LLVM IR generation adds `nounwind` to such functions which makes it UB to unwind out of. IIUC lang team is saying this is deliberately and explicitly UB until explicitly stated to be supported in some cases, for which there's an ongoing RFC, https://github.com/rust-lang/rfcs/pull/2699. The safeguard that aborts when you try to unwind out of an `extern "C"` Rust function did exist but was reverted twice so far because of fallout for software that relied on it despite it being UB, see https://github.com/rust-lang/rust/issues/58794 if you want the messy history and part of the discussion that led to the aforementioned RFC.

src/what-unsafe-does.md
RalfJung: > Certainly though, if you ever do &mut my_union.a, in my understanding of the current reference semantics, you are making a very material claim of having an instance of that type.

Fair. At this point we entered https://github.com/rust-lang/unsafe-code-guidelines/issues/77: does a reference `&mut T` have to point to a "valid" `T` even when the `T` is not actually loaded from memory? The document here for now, conservatively, says "yes", but I actually think we should delay this until the pointer is actually "followed" and a `T` is materialized. But that is a separate discussion.

> inhabitants don't feel like the right abstraction for talking about packing bits? Like you also appealed to the notion of "a list of bytes", which I agree can reasonably include the empty list, but if I ask for a list, you're not allowed to come back with "well, here's no list at all, that's a list right".

That's why I also have a *relation* in the mathematical sense that relates "abstract values" (the view of the world where `bool` has 2 inhabitants, `()` has 1 and `!` has 0) with the way they are represented on the machine (as lists of bytes). If we assume canonical representaiton for a second (which not all types have, but the three we are considering here do), then that means there are exactly 2 byte lists that represent a valid `bool` (both lists have length 1: `[0x00]` and `[0x01]`), there is exactly 1 byte list that represents a valid `()` (the empty list `[]`), and there is no byte list that represents a valid `!`.

So, I think this translates fine even if you switch between the abstract view of a type (the "value" view) and its concrete view (the "byte list" view).

> Question for you: is union { a: !, b: ! } uninstantiable? i.e. is it equivalent to !? In my mind it should be equivalent to !, but it seems like perhaps you would like it to be instantiable?

I personally would prefer for a union to be just a "bag of bytes": its abstract values are all byte lists of appropriate lengths, and the types in the fields don't matter at all (except for their size and alignment) until some field is actually used.

My view of a union of size `N` is basically that is is the same as `[MaybeUninit<u8>; N]`, and that we also offer field access as a convenient way to transmute that type to the type of the field (truncating the data if the field does not cover the entire union).

There are [problems with that view](https://github.com/rust-lang/unsafe-code-guidelines/issues/156), unfortunately. But they are not related to `!`.

So, with that view, the union you are asking about is indeed "instantiable"; it has size 0 and the empty byte list is valid for that type. But accessing any of its fields is a transmute-to-`!` and therefore UB.

> which is strange, because that isn't the definition it uses!

What do you mean by "the definition it uses"?

src/what-unsafe-does.md
```diff
@@ -16,18 +16,41 @@ to your program. You definitely *should not* invoke Undefined Behavior.
 Unlike C, Undefined Behavior is pretty limited in scope in Rust. All the core
 language cares about is preventing the following things:
 
-* Dereferencing null, dangling, or unaligned pointers
-* Reading [uninitialized memory][]
+* Dereferencing (using the `*` operator on) dangling, or unaligned pointers, or
+  wide pointers with invalid metadata (see below)
```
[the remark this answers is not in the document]
RalfJung: So this is just editorial, right? LGTM. I'd prefer if we explained what "dereferencing" means. Specifically, I want to be sure that people consider `&*x` as dereferencing `x`.

> It would be nice if we could use "invalid" here to establish a blanket term for these properties,

Yeah a term would be nice. The problem is that "valid (raw) pointer" would be interpreted as referring to the validity invariant of a raw pointer (defined a bit further down), and that is not the same thing.

This *is*, however, the validity invariant of a reference that we are repeating here. So maybe we could say something like this?

"Dereferencing (using the `*` operator on) a pointer that is not a valid reference (see below)"

src/what-unsafe-does.md
```diff
@@ -16,18 +16,41 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
 * Breaking the [pointer aliasing rules][]
-* Producing invalid primitive values:
-    * dangling/null references
-    * null `fn` pointers
+* Unwinding into another language
+* Causing a [data race][race]
+* Executing code compiled with target features that the current thread of execution does
+  not support (see [`target_feature`][])
+* Producing invalid primitive values (either alone or as a field of a compound
+  type such as `enum`/`struct`/array/tuple):
```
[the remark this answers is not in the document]
RalfJung: The answer is "we don't know yet, see https://github.com/rust-lang/unsafe-code-guidelines/issues/73".

So yes, this is a good question, and one that I would prefer we could skip over for now.

src/what-unsafe-does.md
```diff
@@ -16,18 +16,41 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
-* Causing a [data race][race]
+    * a `!` (all values are invalid for this type)
+    * dangling/unaligned references, references that do themselves point to
+      invalid values, or wide references (to a dynamically sized type) with
+      invalid metadata
+        * slice metadata is invalid if the slice has a total size larger than
+          `isize::MAX` bytes in memory
+        * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
+          `Trait` that matches the actual dynamic trait the reference points to
+    * a non-utf8 `str`
```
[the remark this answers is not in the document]
RalfJung: Or maybe we want to skip this entirely because this is just a library invariant?

src/what-unsafe-does.md
```diff
@@ -16,18 +16,41 @@ to your program. You definitely *should not* invoke Undefined Behavior.
...
+    * an integer (`i*`/`u*`), floating point value (`f*`), or raw pointer read from
+      [uninitialized memory][]
+    * an invalid library type with custom invalid values, such as a `NonNull` or
+      the `NonZero` family of types, that is 0
```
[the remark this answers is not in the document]
RalfJung: I have to say "a type with custom invalid values that is one of those values" sounds rather awkward. I don't have a better proposal either, though.

src/what-unsafe-does.md
RalfJung: I applied most of your suggestions but this one is big enough that it is probably easier to hand the PR off to you. ;) I'd love to do a pass over what you got when you are done, if you don't mind.

I like this new text, as usual in you very pointed style! One comment though:

> Additionally, uninitialized memory is always invalid, so you can't assign it to anything

That's not true for `MaybeUninit`.

RalfJung:
@Gankra thanks for all your feedback!

Btw, have you come to a conclusion on the issue of duplicating the list between here and the reference?

Gankra:
I think it's fine/good for now, as it fits the structure of the book right now, but in the fullness of time that might change.

Issue #150, opened 2019-07-22
Recommend MaybeUninit now that it's stable


est31:
The nomicon still recommends `mem::uninitialized` in the "Unchecked Uninitialized Memory" section.

est31:
cc @RalfJung @steveklabnik

RalfJung:
Cc @Gankro this basically needs incorporating your blog post :)

RalfJung:
I submitted a PR at https://github.com/rust-lang-nursery/nomicon/pull/153.

PR #153, opened 2019-07-30
update uninit section to MaybeUninit

RalfJung:
Fixes https://github.com/rust-lang-nursery/nomicon/issues/150

### DIFF
--- a/src/unchecked-uninit.md
+++ b/src/unchecked-uninit.md
@@ -41,7 +41,7 @@ let x = {
     unsafe { mem::transmute::<_, [Box<u32>; SIZE]>(x) }
 };
 
-println!("{:?}", x);
+dbg!(x);
 ```
 
 This code proceeds in three steps:
--- a/src/unchecked-uninit.md
+++ b/src/unchecked-uninit.md
@@ -110,8 +110,7 @@ arbitrary locations of memory can break things are basically uncountable!
 It's worth noting that you don't need to worry about `ptr::write`-style
 shenanigans with types which don't implement `Drop` or contain `Drop` types,
 because Rust knows not to try to drop them. This is what we relied on in the
-above example. Similarly you should be able to assign to fields of partially
-initialized structs directly if those fields don't contain any `Drop` types.
+above example.
 
 However when working with uninitialized memory you need to be ever-vigilant for
 Rust trying to drop values you make like this before they're fully initialized.
--- a/src/unchecked-uninit.md
+++ b/src/unchecked-uninit.md
@@ -8,25 +8,87 @@ Unfortunately this is pretty rigid, especially if you need to initialize your
 array in a more incremental or dynamic way.
 
 Unsafe Rust gives us a powerful tool to handle this problem:
-[`mem::uninitialized`][uninitialized]. This function pretends to return a value
-when really it does nothing at all. Using it, we can convince Rust that we have
-initialized a variable, allowing us to do trickier things with conditional and
-incremental initialization.
+[`MaybeUninit`]. This type can be used to handle memory that has not been fully
+initialized yet.
 
-Unfortunately, this opens us up to all kinds of problems. Assignment has a
-different meaning to Rust based on whether it believes that a variable is
-initialized or not. If it's believed uninitialized, then Rust will semantically
-just memcopy the bits over the uninitialized ones, and do nothing else. However
-if Rust believes a value to be initialized, it will try to `Drop` the old value!
-Since we've tricked Rust into believing that the value is initialized, we can no
-longer safely use normal assignment.
+With `MaybeUninit`, we can initialize an array element-for-element as follows:
 
-This is also a problem if you're working with a raw system allocator, which
-returns a pointer to uninitialized memory.
+```rust
+use std::mem::{self, MaybeUninit};
 
-To handle this, we must use the [`ptr`] module. In particular, it provides
-three functions that allow us to assign bytes to a location in memory without
-dropping the old value: [`write`], [`copy`], and [`copy_nonoverlapping`].
+// Size of the array is hard-coded but easy to change. This means we can't
+// use [a, b, c] syntax to initialize the array, though!
+const SIZE: usize = 10;
+
+let x = {
+    // Create an uninitialized array of `MaybeUninit`. The `assume_init` is
+    // safe because the type we are claiming to have initialized here is a
+    // bunch of `MaybeUninit`s, which do not require initialization.
+    let mut x: [MaybeUninit<Box<u32>>; SIZE] = unsafe {
+        MaybeUninit::uninit().assume_init()
+    };
+
+    // Dropping a `MaybeUninit` does nothing. Thus using raw pointer
+    // assignment instead of `ptr::write` does not cause the old
+    // uninitialized value to be dropped.
+    // Exception safety is not a concern because Box can't panic
+    for i in 0..SIZE {
+        x[i] = MaybeUninit::new(Box::new(i as u32));
+    }
+
+    // Everything is initialized. Transmute the array to the
+    // initialized type.
+    unsafe { mem::transmute::<_, [Box<u32>; SIZE]>(x) }
+};
+
+println!("{:?}", x);
+```
+
+This code proceeds in three steps:
+
+1. Create an array of `MaybeUninit<T>`. With current stable Rust, we have to use
+   unsafe code for this: we take some uninitialized piece of memory
+   (`MaybeUninit::uninit()`) and claim we have fully initialized it
+   ([`assume_init()`][assume_init]). This seems ridiculous, because we didn't!
+   The reason this is correct is that the array consists itself entirely of
+   `MaybeUninit`, which do not actually require initialization. For most other
+   types, doing `MaybeUninit::uninit().assume_init()` produces an invalid
+   instance of said type, so you got yourself some Undefined Behavior.
+
+2. Initialize the array. The subtle aspect of this is that usually, when we use
+   `=` to assign to a value that the Rust type checker considers to already be
+   initialized (like `x[i]`), the old value stored on the left-hand side gets
+   dropped. This would be a disaster. However, in this case, the type of the
+   left-hand side is `MaybeUninit<Box<u32>>`, and dropping that does not do
+   anything! See below for some more discussion of this `drop` issue.
+
+3. Finally, we have to change the type of our array to remove the
+   `MaybeUninit`. With current stable Rust, this requires a `transmute`.
+   This transmute is legal because in memory, `MaybeUninit<T>` looks the same as `T`.
+
+    However, note that in general, `Container<MaybeUninit<T>>>` does *not* look
+   the same as `Container<T>`! Imagine if `Container` was `Option`, and `T` was
+   `bool`, then `Option<bool>` exploits that `bool` only has two valid values,
+   but `Option<MaybeUninit<bool>>` cannot do that because the `bool` does not
+   have to be initialized.
+
+    So, it depends on `Container` whether transmuting away the `MaybeUninit` is
+   allowed. For arrays, it is (and eventually the standard library will
+   acknowledge that by providing appropriate methods).
+
+It's worth spending a bit more time on the loop in the middle, and in particular
+the assignment operator and its interaction with `drop`.  If we would have
+written something like
+```rust,ignore
+*x[i].as_mut_ptr() = Box::new(i as u32); // WRONG!
+```
+we would actually overwrite a `Box<u32>`, leading to `drop` of uninitialized
+data, which will cause much sadness and pain.
+
+The correct alternative, if for some reason we cannot use `MaybeUninit::new`, is
+to use the [`ptr`] module. In particular, it provides three functions that allow
+us to assign bytes to a location in memory without dropping the old value:
+[`write`], [`copy`], and [`copy_nonoverlapping`].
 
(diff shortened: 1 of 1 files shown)


src/unchecked-uninit.md
```diff
@@ -40,59 +91,32 @@ dropping the old value: [`write`], [`copy`], and [`copy_nonoverlapping`].
...
+arbitrary locations of memory can break things are basically uncountable!
 
 It's worth noting that you don't need to worry about `ptr::write`-style
 shenanigans with types which don't implement `Drop` or contain `Drop` types,
-because Rust knows not to try to drop them. Similarly you should be able to
-assign to fields of partially initialized structs directly if those fields don't
-contain any `Drop` types.
+because Rust knows not to try to drop them. This is what we relied on in the
+above example. Similarly you should be able to assign to fields of partially
+initialized structs directly if those fields don't contain any `Drop` types.
```
Gankra: glad to see this is still legit lol

src/unchecked-uninit.md
```diff
@@ -8,25 +8,76 @@ Unfortunately this is pretty rigid, especially if you need to initialize your
...
+// use [a, b, c] syntax to initialize the array, though!
+const SIZE: usize = 10;
+
+let x = {
+    // Create an uninitialized array of `MaybeUninit`. The `assume_init` is
+    // safe because the type we are claiming to have initialized here is a
+    // bunch of `MaybeUninit`s, which do not require initialization.
+    let mut x: [MaybeUninit<Box<u32>>; SIZE] = unsafe {
+        MaybeUninit::uninit().assume_init()
+    };
```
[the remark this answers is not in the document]
RalfJung: Yeah, the name `assume_init` is kind of bad for this corner case.

I can add more turbofish, but I am not sure what that accomplishes?

src/unchecked-uninit.md
```diff
@@ -8,25 +8,76 @@ Unfortunately this is pretty rigid, especially if you need to initialize your
...
+    // assignment instead of `ptr::write` does not cause the old
+    // uninitialized value to be dropped.
+    // Exception safety is not a concern because Box can't panic
+    for i in 0..SIZE {
+        x[i] = MaybeUninit::new(Box::new(i as u32));
+    }
+
+    // Everything is initialized. Transmute the array to the
+    // initialized type.
+    unsafe { mem::transmute::<_, [Box<u32>; SIZE]>(x) }
```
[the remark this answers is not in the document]
RalfJung: So, what do you propose I change here? Just add more comments saying `DANGER WARNING`?

src/unchecked-uninit.md
```diff
@@ -8,25 +8,76 @@ Unfortunately this is pretty rigid, especially if you need to initialize your
...
+        MaybeUninit::uninit().assume_init()
+    };
+
+    // Dropping a `MaybeUninit` does nothing. Thus using raw pointer
+    // assignment instead of `ptr::write` does not cause the old
+    // uninitialized value to be dropped.
+    // Exception safety is not a concern because Box can't panic
+    for i in 0..SIZE {
+        x[i] = MaybeUninit::new(Box::new(i as u32));
+    }
```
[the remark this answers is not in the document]
RalfJung: I can see either approach work. This one is less verbose, but a bit more subtle.

rustdoc originally had `.write(...)` but was recently changed to what this PR does.

> I imagine you're trying to avoid this because this raises annoying questions about &raw? 

There's no `&raw` trouble here because the `MaybeUninit` is "inside" the array. So no `&mut T` has to ever arise.

Gankra:
(trying email reply might go bad)

re: &raw doesn’t matter — a huge use case for MaybeUninit is building a
struct piece by piece, and i have had to deal with panicked devs who saw
your docs saying this was “impossible” (when really it’s just awkward for
the memory model right now)

RalfJung:
> a huge use case for MaybeUninit is building a
struct piece by piece, and i have had to deal with panicked devs who saw
your docs saying this was “impossible” (when really it’s just awkward for
the memory model right now)

Well, it is impossible in the sense that any attempt of doing it is UB (according to the UB PR that you just landed, even). So, I am sticking to that wording.
There's a good reason that the example here is about arrays, not structs. ;)  For arrays we can avoid the UB.

Gankra:
Yeah. I think the issue is the wording is both slightly too strict and vague, and it makes it sound like we have some sort of notion of constructors having semantic power, like C++. 

Just to clarify, since you're accidentally dodging the implicit question: What's really going on is just that, absent deterministic layout like repr(C), you need either `offsetof` or `&raw` to get a pointer to a field without transiently creating a reference to uninitialized memory, which is currently conservatively considered UB (pending a proper decision). Since we don't have either of those operations, component-wise initialization on repr(rust) types is in limbo.

But with that said, this is one of the few cases where I'm honestly just fine with telling users to go for it (create the transient reference), and doing a push to explain there is now a "right" way to do it later. Getting a pointer to a field is just too important to say you can't do it bluntly like that.

src/unchecked-uninit.md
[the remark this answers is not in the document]
harrysarson: If I may give my 50 cents as someone who hasn't written any unsafe code but followed the stabilisation of `MaybeUninit` with interest:

If someone changed the type annotation of `x` (by editing safe code) the code may still compile (I think in this case it wouldn't) but now the "type we are claiming to have initialised here" could **not** be a bunch of `MaybeUninit`s leading to undefined behaviour. 

I may be barking up the wrong tree here, but this was what I thought the first time I read this code snippet. :)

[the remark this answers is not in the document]
Gankra: no it's not about safety or anything, that's not a huge concern here. just trying to reduce the  "wait what are you doing??" factor of this code (unsuccessfully).

src/unchecked-uninit.md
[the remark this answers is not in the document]
Gankra: Yeah, but we should also have at least one or two paragraphs dedicated to like "hey this is a thing" because it's the most common thing I see people asking about, hoping they can just do it.

Also if your type isn't repr(C), adding MaybeUninit technically can lead to layout randomization for no reason other than the compiler being capricious.

RalfJung:
> Yeah. I think the issue is the wording is both slightly too strict and vague, and it makes it sound like we have some sort of notion of constructors having semantic power, like C++.

Well, constructors "atomically" (not in a concurrency sense) create a new value without intermediate state. It's not a big semantic power but it is meaningful.

> But with that said, this is one of the few cases where I'm honestly just fine with telling users to go for it (create the transient reference), and later doing a push to explain there is now a "right" way to do it. Getting a pointer to a field is just too important to say you can't do it bluntly like that.

Fair, though I personally prefer to make people lobby for `&raw`. ;)
But either way, the document here is only about array initialization -- it was before, and it still is. So I am not sure why we are talking about this?

src/unchecked-uninit.md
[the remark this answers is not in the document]
RalfJung: We could add our local copy of `mem::uninitialized` and call that?

src/unchecked-uninit.md
RalfJung: The "should" is weird though, now that I think about it. Are you or are you not able to do this?

This is basically explaining rustc's static initialization tracking.

src/unchecked-uninit.md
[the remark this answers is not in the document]
RalfJung: `MaybeUninit` is `repr(transparent)` -- it is not entirely clear what that means for unions, but transmute is definitely allowed.

> Yeah, but we should also have at least one or two paragraphs dedicated to like "hey this is a thing" because it's the most common thing I see people asking about, hoping they can just do it.

I don't follow. What is a thing?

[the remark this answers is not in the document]
RalfJung: Pushed something. Let me know what you think.

src/unchecked-uninit.md
```diff
@@ -64,6 +64,17 @@ This code proceeds in three steps:
...
+
+    However, note that in general, `Container<MaybeUninit<T>>>` does *not* look
+   the same as `Container<T>`! Imagine if `Container` was `Option`, and `T` was
+   `bool`, then `Option<bool>` exploits that `bool` only has two valid values,
+   but `Option<MaybeUninit<bool>>` cannot do that because the `bool` does not
+   have to be initialized.
+
+    So, it depends on `Container` whether transmuting away the `MaybeUninit` is
+   allowed. For arrays, it is (and eventually the standard library will
+   acknowledge that by providing an appropriate methods).
```
Gankra: a single space indent is how you do this???? sdasfsdf

Gankra:
```suggestion
   acknowledge that by providing appropriate methods).
```

Gankra:
Apply my one fix and squash and this is good to go

Gankra:
(not having to squash is a reward contributors only get for reviews that go extremely slow and are a big mess :)

src/unchecked-uninit.md
RalfJung: Well, 4-space indent is how you do this. I used to then keep the entire paragraph 4-space indented, but this here IMO looks better when reading the ASCII.

RalfJung:
> (not having to squash is a reward contributors only get for reviews that go extremely slow and are a big mess :)

I actually prefer to squash myself. That way the scripts I use detect the branches (local and remote) as merged and delete them. I have not fond a good way to do that when the PR gets squashed by GitHub.

RalfJung:
> Apply my one fix and squash and this is good to go

Done.

RalfJung:
@Gankra anything else needed to land this?

RalfJung:
@Centril based on [this comment](https://github.com/rust-lang-nursery/nomicon/pull/153#issuecomment-522595408), are you fine with landing this?

src/unchecked-uninit.md
```diff
@@ -8,25 +8,87 @@ Unfortunately this is pretty rigid, especially if you need to initialize your
...
+    // safe because the type we are claiming to have initialized here is a
+    // bunch of `MaybeUninit`s, which do not require initialization.
+    let mut x: [MaybeUninit<Box<u32>>; SIZE] = unsafe {
+        MaybeUninit::uninit().assume_init()
+    };
+
+    // Dropping a `MaybeUninit` does nothing. Thus using raw pointer
+    // assignment instead of `ptr::write` does not cause the old
+    // uninitialized value to be dropped.
+    // Exception safety is not a concern because Box can't panic
```
Centril:
```suggestion
    // Exception safety is not a concern because Box cannot panic.
```

src/unchecked-uninit.md
```diff
@@ -8,25 +8,87 @@ Unfortunately this is pretty rigid, especially if you need to initialize your
...
+    for i in 0..SIZE {
```
Centril: Personally I'd write `for (i, x) in x.iter_mut().enumerate() {` -- seems more "idiomatic Rust" and probably elides some bounds checks / gives LLVM less work to do?

src/unchecked-uninit.md
```diff
@@ -8,25 +8,87 @@ Unfortunately this is pretty rigid, especially if you need to initialize your
...
+        x[i] = MaybeUninit::new(Box::new(i as u32));
+    }
+
+    // Everything is initialized. Transmute the array to the
+    // initialized type.
+    unsafe { mem::transmute::<_, [Box<u32>; SIZE]>(x) }
+};
+
+println!("{:?}", x);
```
Centril:
```suggestion
dbg!(x);
```

src/unchecked-uninit.md
Centril: Wait; this seems to be contradicting https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initializing-a-struct-field-by-field ? What's an example of this working?

src/unchecked-uninit.md
```diff
@@ -8,25 +8,87 @@ Unfortunately this is pretty rigid, especially if you need to initialize your
...
+[`MaybeUninit`]. This type can be used to handle memory that has not been fully
+initialized yet.
+
+With `MaybeUninit`, we can initialize an array element-for-element as follows:
+
+```rust
+use std::mem::{self, MaybeUninit};
+
+// Size of the array is hard-coded but easy to change. This means we can't
+// use [a, b, c] syntax to initialize the array, though!
```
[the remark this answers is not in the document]
RalfJung: Your example is not "easy to change": when I change `S`, I also need to change the initializer.

So, what the text says is true, I think. It probably could be made clearer though.

src/unchecked-uninit.md
RalfJung: `can't` is used throughout the nomicon, much more often than cannot. I'd prefer to keep its conversational style, which this fits with.

src/unchecked-uninit.md
RalfJung: I wanted to stick close to the original code. Also for low-level explanations like these, I sometimes find it helpful to write code more "direct" -- closer to what one would write e.g. in C.

src/unchecked-uninit.md
RalfJung: True, that comment is not really applicable. You have to always initialize through a raw pointer.

Centril:
@RalfJung Did you consider the two additional points re. field-by-field init and casting aspersions on `mem::uninitialized`? (not sure if this is "waiting on author" so to speak)

RalfJung:
I did, but it seems the push failed (likely because I was roaming via mobile at the time). Should be up now.

Centril:
It means I'm waiting on Travis. =P

PR #157, opened 2019-08-14
Update the diagnostic of `error[E0597]` in dropck.md

chansuke:
I was reading [this page](https://github.com/rust-lang-nursery/nomicon/blob/master/src/dropck.md) with [playground(2018)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018) and found that the diagnostic of `error[E0597]` of this example 

```
struct Inspector<'a>(&'a u8);

impl<'a> Drop for Inspector<'a> {
    fn drop(&mut self) {
        println!("I was only {} days from retirement!", self.0);
    }
}

struct World<'a> {
    inspector: Option<Inspector<'a>>,
    days: Box<u8>,
}

fn main() {
    let mut world = World {
        inspector: None,
        days: Box::new(1),
    };
    world.inspector = Some(Inspector(&world.days));
    // Let's say `days` happens to get dropped first.
    // Then when Inspector is dropped, it will try to read free'd memory!
}
```

is slightly different with an output of the current version and it would be better if it can update its error message to the current one?

### DIFF
--- a/src/dropck.md
+++ b/src/dropck.md
@@ -2,8 +2,8 @@
 
 We have seen how lifetimes provide us some fairly simple rules for ensuring
 that we never read dangling references. However up to this point we have only ever
-interacted with the *outlives* relationship in an inclusive manner. That is,
-when we talked about `'a: 'b`, it was ok for `'a` to live *exactly* as long as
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):
Thanks!

PR #161, opened 2019-08-24
Update sample code with stable alloc crate

lzutao:
Based on #170 .

### DIFF
--- a/src/vec-alloc.md
+++ b/src/vec-alloc.md
@@ -15,15 +15,24 @@ want to use `empty` because there's no real allocation to talk about but
 
 So:
 
-```rust,ignore
-#![feature(alloc, heap_api)]
-
+```rust
+# #![feature(ptr_internals)]
+# use std::ptr::{self, Unique};
 use std::mem;
 
+# pub struct Vec<T> {
+#     ptr: Unique<T>,
+#     cap: usize,
+#     len: usize,
+# }
 impl<T> Vec<T> {
-    fn new() -> Self {
+    pub fn new() -> Self {
         assert!(mem::size_of::<T>() != 0, "We're not ready to handle ZSTs");
-        Vec { ptr: Unique::empty(), len: 0, cap: 0 }
+        Vec {
+            ptr: Unique::empty(),
+            len: 0,
+            cap: 0,
+        }
     }
 }
 ```
@@ -37,8 +46,8 @@ that, we'll need to use the rest of the heap APIs. These basically allow us to
 talk directly to Rust's allocator (jemalloc by default).
 
 We'll also need a way to handle out-of-memory (OOM) conditions. The standard
-library calls `std::alloc::oom()`, which in turn calls the the `oom` langitem,
-which aborts the program in a platform-specific manner.
+library calls `std::alloc::handle_alloc_error()`, which in turn calls the the
+`oom` langitem, which aborts the program in a platform-specific manner.
 The reason we abort and don't panic is because unwinding can cause allocations
 to happen, and that seems like a bad thing to do when your allocator just came
 back with "hey I don't have any more memory".
@@ -157,24 +166,22 @@ such we will guard against this case explicitly.
 Ok with all the nonsense out of the way, let's actually allocate some memory:
 
 ```rust,ignore
-use std::alloc::oom;
+use std::alloc;
 
 fn grow(&mut self) {
     // this is all pretty delicate, so let's say it's all unsafe
     unsafe {
-        // current API requires us to specify size and alignment manually.
-        let align = mem::align_of::<T>();
-        let elem_size = mem::size_of::<T>();
+        let layout = alloc::Layout::new::<T>();
 
         let (new_cap, ptr) = if self.cap == 0 {
-            let ptr = heap::allocate(elem_size, align);
+            let ptr = alloc::alloc(layout);
             (1, ptr)
         } else {
             // as an invariant, we can assume that `self.cap < isize::MAX`,
             // so this doesn't need to be checked.
             let new_cap = self.cap * 2;
             // Similarly this can't overflow due to previously allocating this
-            let old_num_bytes = self.cap * elem_size;
+            let old_num_bytes = self.cap * mem::size_of::<T>();
 
             // check that the new allocation doesn't exceed `isize::MAX` at all
             // regardless of the actual size of the capacity. This combines the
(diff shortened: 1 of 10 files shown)


src/vec-alloc.md
```diff
@@ -157,46 +166,45 @@ such we will guard against this case explicitly.
...
+                old_num_bytes <= (std::isize::MAX as usize) / 2,
+                "capacity overflow"
+            );
 
             let new_num_bytes = old_num_bytes * 2;
-            let ptr = heap::reallocate(self.ptr.as_ptr() as *mut _,
-                                        old_num_bytes,
-                                        new_num_bytes,
-                                        align);
+            let ptr = alloc::realloc(self.ptr.as_ptr() as *mut u8, layout, new_num_bytes);
```
Gankra: This isn't correct. The layout you need to pass here is for an array, but this is the layout for a single value. You must use from_size_align_unchecked or from_size_align (generally the former is more appropriate since we're doing our own error-handling).

src/vec-insert-remove.md
```diff
@@ -16,17 +16,20 @@ using the old len.
 pub fn insert(&mut self, index: usize, elem: T) {
     // Note: `<=` because it's valid to insert after everything
     // which would be equivalent to push.
-    assert!(index <= self.len, "index out of bounds");
-    if self.cap == self.len { self.grow(); }
+    assert!(self.len >= index, "index is out of bounds");
```
Gankra: Why did you invert this assertion? This is not standard style.

src/vec-insert-remove.md
```diff
@@ -16,17 +16,20 @@ using the old len.
...
         if index < self.len {
-            // ptr::copy(src, dest, len): "copy from source to dest len elems"
-            ptr::copy(self.ptr.offset(index as isize),
-                      self.ptr.offset(index as isize + 1),
-                      self.len - index);
+            ptr::copy(
+                self.ptr.as_ptr().offset(index as isize),
+                self.ptr.as_ptr().offset(index as isize + 1),
+                self.len - index,
+            );
```
Gankra: Why did you remove this comment?

src/vec-insert-remove.md
```diff
@@ -38,13 +41,18 @@ Remove behaves in the opposite manner. We need to shift all the elements from
 ```rust,ignore
 pub fn remove(&mut self, index: usize) -> T {
     // Note: `<` because it's *not* valid to remove after everything
-    assert!(index < self.len, "index out of bounds");
+    assert!(self.len > index, "index is out of bounds");
+
```
Gankra: Another changed assertion.

src/vec-insert-remove.md
```diff
@@ -38,13 +41,18 @@ Remove behaves in the opposite manner. We need to shift all the elements from
...
-                  self.ptr.offset(index as isize),
-                  self.len - index);
+        let result = ptr::read(self.ptr.as_ptr().offset(index as isize));
+
+        ptr::copy(
+            self.ptr.as_ptr().offset(index as isize + 1),
+            self.ptr.as_ptr().offset(index as isize),
+            self.len - index,
+        );
+
```
Gankra: Why are you reformatting the code?

src/vec-into-iter.md
```diff
@@ -134,12 +133,8 @@ impl<T> Drop for IntoIter<T> {
         if self.cap != 0 {
             // drop any remaining elements
             for _ in &mut *self {}
-
-            let align = mem::align_of::<T>();
-            let elem_size = mem::size_of::<T>();
-            let num_bytes = elem_size * self.cap;
             unsafe {
-                heap::deallocate(self.buf.as_ptr() as *mut _, num_bytes, align);
+                alloc::dealloc(self.buf.as_ptr() as *mut _, alloc::Layout::new::<T>());
```
Gankra: Another place with an incorrect Layout.

src/vec-raw.md
```diff
@@ -18,45 +18,47 @@ struct RawVec<T> {
...
-
 impl<T> Drop for RawVec<T> {
     fn drop(&mut self) {
         if self.cap != 0 {
-            let align = mem::align_of::<T>();
-            let elem_size = mem::size_of::<T>();
-            let num_bytes = elem_size * self.cap;
             unsafe {
-                heap::deallocate(self.ptr.as_mut() as *mut _, num_bytes, align);
+                alloc::dealloc(self.ptr.as_ptr() as *mut u8, alloc::Layout::new::<T>());
```
Gankra: Another incorrect Layout

src/vec-raw.md
```diff
@@ -116,15 +118,20 @@ impl<T> Drop for IntoIter<T> {
 impl<T> Vec<T> {
     pub fn into_iter(self) -> IntoIter<T> {
         unsafe {
-            // need to use ptr::read to unsafely move the buf out since it's
+            // need to use `ptr::read` to unsafely move the buf out since it's
```
Gankra: These backticks aren't really helpful inside comments, since the markdown is never rendered.

src/vec-raw.md
```diff
@@ -116,15 +118,20 @@ impl<T> Drop for IntoIter<T> {
...
             IntoIter {
-                start: *buf.ptr,
-                end: buf.ptr.offset(len as isize),
+                start: buf.ptr.as_ptr(),
+                end: if buf.cap == 0 {
+                    // can't offset off this pointer, it's not allocated!
+                    buf.ptr.as_ptr()
+                } else {
+                    buf.ptr.as_ptr().offset(len as isize)
+                },
```
Gankra: I'm pretty sure this is unnecessary, but it's fair this isn't documented properly.

https://github.com/rust-lang/rust/issues/65108

src/vec-zsts.md
```diff
@@ -35,55 +35,57 @@ method of RawVec.
...
 
     fn grow(&mut self) {
         unsafe {
-            let elem_size = mem::size_of::<T>();
-
-            // since we set the capacity to usize::MAX when elem_size is
+            // Since we set the capacity to `usize::MAX` for ZST is
             // 0, getting to here necessarily means the Vec is overfull.
-            assert!(elem_size != 0, "capacity overflow");
+            assert!(mem::size_of::<T>() != 0, "capacity overflow");
```
Gankra: You should make this check the layout.size(), but I guess that doesn't matter because you probably need to redo this code because this is the wrong layout in the general case.

src/vec-zsts.md
```diff
@@ -35,55 +35,57 @@ method of RawVec.
...
             } else {
-                let new_cap = 2 * self.cap;
-                let ptr = heap::reallocate(self.ptr.as_ptr() as *mut _,
-                                            self.cap * elem_size,
-                                            new_cap * elem_size,
-                                            align);
+                let new_cap = self.cap * 2;
+                let ptr = alloc::realloc(
+                    self.ptr.as_ptr() as *mut u8,
+                    layout,
```
Gankra: Another wrong layout.

src/vec-zsts.md
```diff
@@ -35,55 +35,57 @@ method of RawVec.
...
-        // don't free zero-sized allocations, as they were never allocated.
-        if self.cap != 0 && elem_size != 0 {
-            let align = mem::align_of::<T>();
-
-            let num_bytes = elem_size * self.cap;
+        // Do not free zero-sized allocations, as they were never allocated.
+        if self.cap != 0 && mem::size_of::<T>() != 0 {
             unsafe {
-                heap::deallocate(self.ptr.as_ptr() as *mut _, num_bytes, align);
+                alloc::dealloc(self.ptr.as_ptr() as *mut u8, alloc::Layout::new::<T>());
```
Gankra: Another incorrect layout.

CreepySkeleton:
@lzutao could you please also replace `Unique` with `NonNull` so it's finally works on stable?

lzutao:
@CreepySkeleton Thanks for the interest. 
I don't think NonNull could replace Unique. Could you elaborate more about that?
If you want to read about Unique, here is the link: https://doc.rust-lang.org/nightly/src/core/ptr/unique.rs.html#8

CreepySkeleton:
@lzutao 
What's similar between `NonNull` and `Unique`
- They both cannot be `null`
- Both are pointers
- Both are covariant over `T`

`Unique`'s features `NonNull` lacks:
- `notNull` is neither `Send` nor `Sync` because it can be aliased. But the implementation guarantees that the pointer won't ever be aliased so I believe we can just do this
  ```rust
  unsafe impl<T: Send> Send for Vec<T> {}
  unsafe impl<T: Sync> Sync for Vec<T> {}
```

*AFAIK*, that's all. As we can see, in this case `Unique` *can* be replaced by `NonNull` very well.

clifton:
ill close #182 as its a duplicate (although just covering all aspects of the vec example) but may be worth looking to see if theres anything worth adding from there.

it would be good to merge in something like this, as i was not able to get any of the examples to compile on nightly and im sure others have the same problem.

lzutao:
I have no interest in this PR.

PR #162, opened 2019-08-25
adjust for current reality wrt. wide raw pointers

RalfJung:
Adjust for https://github.com/rust-lang/rust/issues/63851

### DIFF
--- a/src/what-unsafe-does.md
+++ b/src/what-unsafe-does.md
@@ -16,8 +16,7 @@ to your program. You definitely *should not* invoke Undefined Behavior.
 Unlike C, Undefined Behavior is pretty limited in scope in Rust. All the core
 language cares about is preventing the following things:
 
-* Dereferencing (using the `*` operator on) dangling, or unaligned pointers, or
-  wide pointers with invalid metadata (see below)
+* Dereferencing (using the `*` operator on) dangling or unaligned pointers (see below)
 * Breaking the [pointer aliasing rules][]
 * Unwinding into another language
 * Causing a [data race][race]
@@ -36,6 +35,7 @@ language cares about is preventing the following things:
           `isize::MAX` bytes in memory
         * `dyn Trait` metadata is invalid if it is not a pointer to a vtable for
           `Trait` that matches the actual dynamic trait the reference points to
+    * a wide raw pointer that has invalid metadata (see above)
     * a `str` that isn't valid UTF-8
     * an integer (`i*`/`u*`), floating point value (`f*`), or raw pointer read from
       [uninitialized memory][]


src/what-unsafe-does.md
```diff
@@ -30,8 +29,8 @@ language cares about is preventing the following things:
     * a null `fn` pointer
     * a `char` outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
     * a `!` (all values are invalid for this type)
-    * a reference that is dangling, unaligned, points to an invalid value, or
-      that has invalid metadata (if wide)
+    * a reference/`Box` that is dangling, unaligned, or points to an invalid value.
+    * a wide reference, `Box` or raw pointer that has invalid metadata:
```
Centril:
```suggestion
    * a wide reference, `Box`, or raw pointer that has invalid metadata:
```

RalfJung:
Turns out that for raw slices, the length condition can be [violated in safe code](https://github.com/rust-lang/rust/issues/63851#issuecomment-524723952).

So maybe I should just remove the slice condition here, effectively making it part of the safety invariant but not the validity invariant? That would preclude layout optimizations but otherwise not make much of a difference.

RalfJung:
I fixed the slice situation.

RalfJung:
@gnzlbg see above. I used your wording, there are still open questions.

gnzlbg:
> I am a bit confused: is it really UB to just call a function that you think is no-unwind, but actually can unwind?

I think this works in practice for most unwinding ABIs, I don't know if it works for all. Does any unwinding ABI require caller cooperation even when it does not unwind? (cc @comex ?). 

> Like, you don't need to actually unwind for it to be UB?

Due to something like https://github.com/rust-lang/unsafe-code-guidelines/issues/198 one doesn't even need to call it for it to be UB. For example, see https://rust.godbolt.org/z/eHY8dO

```rust
#![feature(unwind_attributes)]
pub mod bad0 {
    extern "Rust" { fn foo(); }
    pub unsafe fn call() -> usize { foo as usize }
}

pub mod bad1 {
    extern "Rust" { #[unwind(allow)]  fn foo(); }
    pub unsafe fn call() { foo() } // UB because `foo` is nounwind because bad0::foo is nounwind
}
```

We can probably fix that, by never making extern declarations `nounwind`, but..

src/what-unsafe-does.md
```diff
@@ -16,10 +16,9 @@ to your program. You definitely *should not* invoke Undefined Behavior.
 Unlike C, Undefined Behavior is pretty limited in scope in Rust. All the core
 language cares about is preventing the following things:
 
-* Dereferencing (using the `*` operator on) dangling, or unaligned pointers, or
-  wide pointers with invalid metadata (see below)
+* Dereferencing (using the `*` operator on) dangling or unaligned pointers (see below)
 * Breaking the [pointer aliasing rules][]
-* Unwinding into another language
+* Calling a function with the wrong call ABI (in particular, with the wrong unwind ABI).
```
gnzlbg:
```suggestion
* Calling a function with the wrong call ABI or wrong unwind ABI.
```

RalfJung:
We would likely want to disallow `extern { ... }` declarations without explicit `unwind` for such ABIs -- just like we disallow `extern { ... }` declarations without argument types.

The current `extern { ... }` syntax is really only suited for ABIs where there is a "default" for the unwinding ABI that always works, assuming unwinding doesn't happen.

gnzlbg:
> since those would be new calling conventions, we can add define that UB if they ever come into existence, right?

Maybe? The problem is that `extern "C"` and `extern "system"` are stable and follow the platform calling convention, so if we were to support a platform where that happened, we'd need to do that for these ABIs there. We could just say that we'd never support those platforms and call it a day. As @comex says, it's unlikely to happen. 

Still, for the wording here, does changing "call ABI or unwinding ABI" to "call ABI" or even just "ABI" make a difference? We haven't written down anywhere whether the unwinding ABI is part of the call ABI or not, and it is arguably part of the "ABI" in general...

---

@comex 

>  I remember someone (maybe even you?) proposing that Rust could invent its own portable unwinding mechanism based on effectively wrapping all function return values in Result; that would probably require caller cooperation.

Are you referring to: https://github.com/CraneStation/cranelift/issues/553 ?

RalfJung:
> Still, for the wording here, does changing "call ABI or unwinding ABI" to "call ABI" or even just "ABI" make a difference? We haven't written down anywhere whether the unwinding ABI is part of the call ABI or not, and it is arguably part of the "ABI" in general...

In terms of wording I would propose something like

> Calling a function with the wrong call ABI or unwinding from a function with the wrong unwind ABI.

RalfJung:
FWIW, the corresponding Reference PR landed (which is in sync with this one).

Gankra:
PSA i am dealing with medecine issues atm and am completely out of my mind and in too much pain to think for the forseeable

gnzlbg:
@Gankra I hope you get well soon, take care!

RalfJung:
@Gankra all the best!

---

So how to we handle the nomicon in the mean time? Are there other editors? @Centril do you know? Are you fine landing this PR, to get Nomicon and reference in sync (now that we merged https://github.com/rust-lang-nursery/reference/pull/659)?

Centril:
I don't have r+ in this repo myself but I think the reference editors + UCG + T-lang should have r+ here.

That said I think I'd need to actually review the PR first myself. ;)

src/what-unsafe-does.md
```diff
@@ -51,8 +50,10 @@ points to are part of the same allocation (so in particular they all have to be
 part of *some* allocation). The span of bytes it points to is determined by the
 pointer value and the size of the pointee type. As a consequence, if the span is
 empty, "dangling" is the same as "non-null". Note that slices point to their
-entire range, so it's very important that the length metadata is never too
-large. If for some reason this is too cumbersome, consider using raw pointers.
+entire range, so it's very important that the length metadata is never too large
```
Centril:
```suggestion
entire range, so it's important that the length metadata is never too large
```

RalfJung: FWIW I think "very" is much more in line with the style of the Nomicon. But I am too lazy to argue...

RalfJung:
> I don't have r+ in this repo myself but I think the reference editors + UCG + T-lang should have r+ here.

Reference editors + T-lang sounds reasonable; UCG doesn't really have a formal notion of "membership" but I guess we could suggest some people to T-lang and see if T-lang is happy giving them r+ for the Nomicon.

Which team is the one to make such decisions?

gnzlbg:
@pietroalbini :

> I don't have r+ in this repo myself but I think the reference editors + UCG + T-lang should have r+ here.

Could you give invite T-lang as maintainers of this repo, or at least @Centril ?

pietroalbini:
I can make the changes to permissions, yes, but I shouldn't be the one deciding those.

Centril:
@pietroalbini As a start, can you grant the lang-team members, ehuss, matthewjasper, and RalfJung r+?

pietroalbini:
@Centril have we considered the fact the lang team doesn't exist on the nursery? :upside_down_face:

Gankra:
Just want to throw my hat out there again that I think the nomicon should be in the rust-lang/rust repo so that unstable code examples can be atomically adjusted along with the APIs.

PR #167, opened 2019-10-09
Fix the versioning of libc

chansuke:
The following versioning 
`libc = "0.2"` 
should be better? 

ref:[crate.io](https://crates.io/crates/libc), [README](https://github.com/rust-lang/libc)

### DIFF
--- a/src/ffi.md
+++ b/src/ffi.md
@@ -16,7 +16,7 @@ examples yourself, you’ll need to add `libc` to your `Cargo.toml`:
 
 ```toml
 [dependencies]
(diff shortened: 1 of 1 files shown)


Gankra:
@steveklabnik any opinion from the SemVer guru who also wrote this section?

steveklabnik:
I personally prefer to write the whole thing out, but strictly speaking, these two things are equivalent. I've seen a lot of both in the ecosystem.

Gankra:
yeah i like the full version too, but thanks for the patch!

PR #169, opened 2019-10-12
further clarify C11 and C/C++11 terminology

RalfJung:
Thanks @petrochenkov for confirming that C++ is the origin of this model.


RalfJung:
> Question: does the reference say anything that needs to be updated as well?

I haven't read most of it, so no idea.^^

`rg C11` gives no results.

Centril:
> `rg C11` gives no results.

Neither does "Concurrency", "Atomic", ... :)

Centril:
The book doesn't seem to mention this at all so we're good there.

PR #175, opened 2019-11-20
Update unsafe-code-guidelines link

JohnTitor:
The old link returns 404, we should point to the new place.

### DIFF
--- a/src/other-reprs.md
+++ b/src/other-reprs.md
@@ -1,7 +1,7 @@
 # Alternative representations
 
 Rust allows you to specify alternative data layout strategies from the default.
-There's also the [reference].
+There's also the [unsafe code guidelines] (note that it's **NOT** normative).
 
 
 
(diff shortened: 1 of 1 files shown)


src/other-reprs.md
```diff
@@ -143,7 +143,7 @@ This is a modifier on `repr(C)` and `repr(rust)`. It is incompatible with
 
 
 
-[reference]: https://github.com/rust-rfcs/unsafe-code-guidelines/tree/master/reference/src/representation.md
+[reference]: https://rust-lang.github.io/unsafe-code-guidelines/layout.html
```
Centril: I'm confused why this is referring to the UCG rather than the actual reference...
Would be good to change the text itself to say "unsafe code guidelines" and then explicitly note that it's also not normative.

JohnTitor: Indeed. Updated the sentence also.

Centril:
Thanks!

PR #177, opened 2019-12-12
Remove mention of contravariance possibly getting scrapped

petertodd:
Contravariance is part of stable Rust, and it's a natural outcome of how `fn(T)` interacts with lifetimes, so it's hard to imagine this ever going away.

tl;dr: LONG LIVE CONTRAVARIANCE!


petertodd:
I'll have to admit an ulterior motive here though: I actually have a project with a `PhantomData<fn(&'a ())>` in one of the types, because the API needs contravariance over `'a`. So I'd really like it to not go away. :)

Centril:
cc @matthewjasper @nikomatsakis

matthewjasper:
Thanks!

nikomatsakis:
I agree we will not be removing it

PR #181, opened 2019-12-30
Add note about using nightly rust in vec tutorial

clifton:
This should save future readers some time.


clifton:
closing in favor of #181

PR #182, opened 2019-12-30
Update vec example to 1.42 nightly

clifton:
Everything necessary to get the Vec example working in 1.42 nightly.


clifton:
closed in favor of #161

PR #186, opened 2020-01-24
Update send-and-sync.md

ChrisGreenaway:
I think this line is trying to give &T as an example of something that is Sync.


Gankra:
No, it's trying to say something is `T: Sync` iff `&T: Send`

PR #204, opened 2020-03-29
Delete redundant & unprof. lang from transmutes

Pand9:
These three points don't add anything and they don't belong to professional documentation.


Gankra:
Based on past history this point necessitates this level of emphasis, and the slightly silly nature of the emphasis makes it more memorable. The rustonomicon is not required or expected to be 100% dry reading.

PR #206, opened 2020-04-03
Update vec-alloc.md to no longer claim jemalloc backs the Rust heap APIs by default

### LINKED ISSUE #229: Incorrect parenthetical statement that jemalloc is the default allocator
It seems like the [nomicon states that jemalloc is the default allocator](https://doc.rust-lang.org/nomicon/vec-alloc.html), however this is not true anymore (Rust uses the system allocator by deafult now).

> Next we need to figure out what to actually do when we do want space. For that, we'll need to use the rest of the heap APIs. These basically allow us to talk directly to Rust's allocator **(jemalloc by default)**.

alex:
Fixes #229

### DIFF
--- a/src/vec-alloc.md
+++ b/src/vec-alloc.md
@@ -34,7 +34,7 @@ Without this assert, some of our early drafts will do some Very Bad Things.
 
 Next we need to figure out what to actually do when we *do* want space. For
 that, we'll need to use the rest of the heap APIs. These basically allow us to
-talk directly to Rust's allocator (jemalloc by default).
+talk directly to Rust's allocator (libc's `malloc` by default).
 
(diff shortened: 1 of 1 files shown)


alex:
Test failures look unrelated to this PR.

src/vec-alloc.md
```diff
@@ -34,7 +34,7 @@ Without this assert, some of our early drafts will do some Very Bad Things.
 
 Next we need to figure out what to actually do when we *do* want space. For
 that, we'll need to use the rest of the heap APIs. These basically allow us to
-talk directly to Rust's allocator (jemalloc by default).
+talk directly to Rust's allocator (libc's `malloc` by default).
```
JohnTitor, line 37: We could also mention `HeapAlloc`?

PR #207, opened 2020-04-05
Use isize::MAX directly on type instead of module

faern:
It's simpler (shorter) to access `MIN`/`MAX` directly on the integer types now. This repository only had a single instance of this, so the PR is tiny. But it's basically a follow up of https://github.com/rust-lang/rust/pull/69860 to make all documentation consistently use the newest versions of the constants.

r? @dtolnay

### DIFF
--- a/src/vec-alloc.md
+++ b/src/vec-alloc.md
@@ -182,7 +182,7 @@ fn grow(&mut self) {
             // we need to make. We lose the ability to allocate e.g. 2/3rds of
             // the address space with a single Vec of i16's on 32-bit though.
             // Alas, poor Yorick -- I knew him, Horatio.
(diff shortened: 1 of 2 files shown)


Centril:
Oh, wrong repo... no bors here :P

faern:
Oops. I pushed a fix for what the CI failed at at basically the same time as you approved. But it only fixed one of two errors. Should I try to fix the remaining one, merge as is, or revert that fix?

Centril:
We'll need to fix the CI errors before merging.

faern:
The errors were not caused by this PR. But we can fix them while at it, sure :)

faern:
@Centril Done!

PR #218, opened 2020-05-16
fix Nomicon transmute UB

RalfJung:
The fact that the Nomicon says "Transmuting between non-repr(C) types is UB" [regularly leads to confusion](https://github.com/rust-lang/nomicon/issues/6#issuecomment-553832993), so let's fix that.


jamesmunns:
This helps clarify things a lot, thanks @RalfJung!

RalfJung:
Cc @rust-lang/lang

nikomatsakis:
Seems good to me.

PR #219, opened 2020-06-03
Data Race definition should be more precise

isakcodes:
"one of them is a write" did not seem right to me. Doesn't generalise to the case with two concurrent threads writing to shared memory. [Related Wikipedia article](https://en.wikipedia.org/wiki/Race_condition#Example_definitions_of_data_races_in_particular_concurrency_models)


JohnTitor (approved):
Looks good, thanks!

PR #221, opened 2020-06-05
Explain additional rules for `unsafe` blocks in `const` and `const fn`

oli-obk:
r? @RalfJung

### DIFF
--- a/src/const-safety.md
+++ b/src/const-safety.md
@@ -1,0 +1,171 @@
+# Const safety
+
+The miri engine, which is used to execute code at compile time, can fail in
+four possible ways:
+
+* The program performs an unsupported operation (e.g., calling an unimplemented
+  intrinsics, or doing an operation that would observe the integer address of a
+  pointer).
+* The program causes undefined behavior (e.g., dereferencing an out-of-bounds
+  pointer).
+* The program panics (e.g., a failed bounds check).
+* The program exhausts its resources: It might overflow the stack, allocation
+  too much memory or loops forever.  Note that detecting these conditions
+  happens on a best-effort basis only.
+
+Just like panics and non-termination are acceptable in safe run-time Rust code,
+we also consider these acceptable in safe compile-time Rust code.  However, we
+would like to rule out the first two kinds of failures in safe code.  Following
+the terminology in [this blog post], we call a program that does not fail in the
+first two ways *const safe*.
+
+[this blog post]: https://www.ralfj.de/blog/2018/07/19/const.html
+
+The goal of the const safety check, then, is to ensure that a program is const
+safe.  What makes this tricky is that there are some operations that are safe as
+far as run-time Rust is concerned, but are inherently unsupportable in the miri engine and hence
+not const safe (they fall in the first category of failures above).
+We call these operations *unconst*.  The purpose
+of the following section is to explain this in more detail, before proceeding
+with the main definitions.
+
+## Miri background
+
+A very simple example of an unconst operation is
+
+```rust
+static S:i32 = 0;
+const BAD:bool = (&S as *const i32 as usize) % 16 == 0;
+```
+
+The modulo operation here is not supported by the miri engine because evaluating
+it requires knowing the actual integer address of `S`.
+
+The way miri handles this is by treating pointer and integer values separately.
+The most primitive kind of value in miri is a `Scalar`, and a scalar is *either*
+a pointer (`Scalar::Ptr`) or a bunch of bits representing an integer
+(`Scalar::Bits`).  Every value of a variable of primitive type is stored as a
+`Scalar`.  In the code above, casting the pointer `&S` to `*const i32` and then
+to `usize` does not actually change the value -- we end up with a local variable
+of type `usize` whose value is a `Scalar::Ptr`.  This is not a problem in
+itself, but then executing `%` on this *pointer value* is unsupported.
+
+However, it does not seem appropriate to blame the `%` operation above for this
+failure. `%` on "normal" `usize` values (`Scalar::Bits`) is perfectly fine, just using it on
+values computed from pointers is an issue.  Essentially, `&i32 as *const i32 as
+usize` is a "safe" `usize` at run-time (meaning that applying safe operations to
+this `usize` cannot lead to misbehavior, following terminology [suggested here])
+-- but the same value is *not* "safe" at compile-time, because we can cause a
+const safety violation by applying a safe operation (namely, `%`).
+
+This is similar to how it is illegal in Rust to have a dangling reference, because you
+can safely dereference it, which would be undefined behavior. Even though the issue
+occurs at the time of the dereference, we treat the creation of a dangling reference
+as the unsound action, not the final dereference.
+
+[suggested here]: https://www.ralfj.de/blog/2018/08/22/two-kinds-of-invariants.html
+
+## Const safety check on values
+
+The result of any const computation (`const`, `static`, promoteds) is subject to
+a "sanity check" which enforces const safety.  Const safety
+is defined as follows:
+
+* Integer and floating point types are const-safe if they are a `Scalar::Bits`.
+  This makes sure that we can run `%` and other operations without violating
+  const safety.  In particular, the value must *not* be uninitialized.
+* References are const-safe if they are `Scalar::Ptr` into allocated memory, and
+  the data stored there is const-safe.  (Technically, we would also like to
+  require `&mut` to be unique and the memory behind a `&` to not be mutable unless there is an
+  `UnsafeCell`, but it seems infeasible to check that.)  For wide pointers, the
+  length of a slice must be a valid `usize` and the vtable of a `dyn Trait` must
+  be a valid vtable.
+* `bool` is const-safe if it is `Scalar::Bits` with a value of `0` or `1`.
+* `char` is const-safe if it is a valid unicode codepoint.
+* `()` is always const-safe.
+* `!` is never const-safe.
+* `dyn Trait` is const-safe if the value is const-safe at the type indicated by
+  the vtable.
+* Function pointers are const-safe if they point to an actual function.  A
+  `const fn` pointer (when/if we have those) must point to a `const fn`.
+* Raw pointers are const safe if they are either `Scalar::Bits` or are not dangling
+* Aggregates must not contain dangling `Scalar::Ptr` (e.g. in padding or unions) and additionally
+  follow the following aggregate-specific rules:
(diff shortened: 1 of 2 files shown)


src/const-safety.md
```diff
@@ -0,0 +1,171 @@
+# Const safety
+
+The miri engine, which is used to execute code at compile time, can fail in
```
RalfJung, line 3: miri or Miri? I have often been writing the latter, but not consistently...^^

RalfJung:
> this is taken in large parts from https://github.com/rust-lang/const-eval/blob/master/const_safety.md

To avoid duplicating material, we should probably make that a link to the Nomicon after landing this.

src/const-safety.md
```diff
@@ -0,0 +1,171 @@
...
+of the following section is to explain this in more detail, before proceeding
+with the main definitions.
+
+## Miri background
+
+A very simple example of an unconst operation is
+
+```rust
+static S:i32 = 0;
+const BAD:bool = (&S as *const i32 as usize) % 16 == 0;
```
RalfJung, line 38:
```suggestion
static S: i32 = 0;
const BAD: bool = (&S as *const i32 as usize) % 16 == 0;
```

src/const-safety.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+The modulo operation here is not supported by the miri engine because evaluating
+it requires knowing the actual integer address of `S`.
+
+The way miri handles this is by treating pointer and integer values separately.
+The most primitive kind of value in miri is a `Scalar`, and a scalar is *either*
+a pointer (`Scalar::Ptr`) or a bunch of bits representing an integer
```
RalfJung, line 46:
```suggestion
a pointer (`Scalar::Ptr`) *or* a bunch of bits representing an integer
```

src/const-safety.md
```diff
@@ -0,0 +1,171 @@
...
+(`Scalar::Bits`).  Every value of a variable of primitive type is stored as a
+`Scalar`.  In the code above, casting the pointer `&S` to `*const i32` and then
+to `usize` does not actually change the value -- we end up with a local variable
+of type `usize` whose value is a `Scalar::Ptr`.  This is not a problem in
+itself, but then executing `%` on this *pointer value* is unsupported.
+
+However, it does not seem appropriate to blame the `%` operation above for this
+failure. `%` on "normal" `usize` values (`Scalar::Bits`) is perfectly fine, just using it on
+values computed from pointers is an issue.  Essentially, `&i32 as *const i32 as
+usize` is a "safe" `usize` at run-time (meaning that applying safe operations to
```
RalfJung, line 56:
```suggestion
usize` computes a "safe" `usize` at run-time (meaning that applying safe operations to
```

src/const-safety.md
oli-obk: I've fairly consistently used the lower case form. :shrug: I have no opinion on it beyond "my fingers like not using more modifier keys than they need to".

src/const-safety.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Const safety check on values
+
+The result of any const computation (`const`, `static`, promoteds) is subject to
+a "sanity check" which enforces const safety.  Const safety
+is defined as follows:
+
+* Integer and floating point types are const-safe if they are a `Scalar::Bits`.
+  This makes sure that we can run `%` and other operations without violating
+  const safety.  In particular, the value must *not* be uninitialized.
```
RalfJung, line 76:
```suggestion
  const safety.  In particular, and just like at run-time, the value must *not* be uninitialized.
```

src/const-safety.md
```diff
@@ -0,0 +1,171 @@
...
+const BAD:bool = (&S as *const i32 as usize) % 16 == 0;
+```
+
+The modulo operation here is not supported by the miri engine because evaluating
+it requires knowing the actual integer address of `S`.
+
+The way miri handles this is by treating pointer and integer values separately.
+The most primitive kind of value in miri is a `Scalar`, and a scalar is *either*
+a pointer (`Scalar::Ptr`) or a bunch of bits representing an integer
+(`Scalar::Bits`).  Every value of a variable of primitive type is stored as a
```
RalfJung, line 47:
```suggestion
(`Scalar::Bits`), or uninitialized.  Every value of a variable of primitive type is stored as a
```
This is mentioned later so might be good to have that here for completeness' sake.

We should likely replace `Scalar` by just the work "scalar" then, and generally talk about concepts rather than specific rustc implementation types.

src/const-safety.md
```diff
@@ -0,0 +1,171 @@
...
+a "sanity check" which enforces const safety.  Const safety
+is defined as follows:
+
+* Integer and floating point types are const-safe if they are a `Scalar::Bits`.
+  This makes sure that we can run `%` and other operations without violating
+  const safety.  In particular, the value must *not* be uninitialized.
+* References are const-safe if they are `Scalar::Ptr` into allocated memory, and
+  the data stored there is const-safe.  (Technically, we would also like to
+  require `&mut` to be unique and the memory behind a `&` to not be mutable unless there is an
+  `UnsafeCell`, but it seems infeasible to check that.)  For wide pointers, the
```
RalfJung, line 80: We should make it clear that, similar to run-time, having aliasing mutable reference at const-time is still wrong even if the sanity check does not catch that. Likely the header above the list should make it clear that the sanity check *approximates* const safety but may actually accept some const-unsafe values.

RalfJung: Alternatively... I don't think we have such a list for normal run-time safety, I think. So this is probably a level of detail that we do not yet want to commit to. Maybe it is better to remove this here and leave the discussion more high-level -- and only keep this level of detail in our const-eval writeup.

src/const-safety.md
```diff
@@ -0,0 +1,171 @@
...
+* inspecting the bits of raw pointers (e.g. by hashing them).
+
+
+These operations are unconst because a `const fn` that
+(when called with the same arguments at runtime and compile-time)
+
+  * fails to run at compile time when it works perfectly fine at runtime
+  * produces a different result at compile time than at runtime
+
+is undefined behavior.
```
RalfJung, line 138: That would be a new kind of UB to be [added to the reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html), and to be checked by Miri. I think the latter will be really hard to do so I am not sure if it's worth it.

src/const-safety.md
```diff
@@ -0,0 +1,171 @@
...
+On the other hand, the following function is *not* const-safe and hence it is considered a bug to mark it as such:
+
+```
+const fn ptr_eq<T>(x: &T, y: &T) -> bool {
+    unconst { x as *const T == y as *const T }
+}
+```
+
+If the function were invoked as `ptr_eq(&42, &42)` the result depends on the potential
+deduplication of the memory of the `42`s.
```
RalfJung, line 171: Or rather, the result is that we get an error of the first kind: an unsupported operation.

src/what-unsafe-does.md
```diff
@@ -7,6 +7,8 @@ The only things that are different in Unsafe Rust are that you can:
 * Implement `unsafe` traits
 * Mutate statics
 * Access fields of `union`s
+* Cast raw pointers to integers within const contexts
+* Compare raw pointers within const contexts
```
RalfJung, line 11: With the `guaranteed_*` intrinsics, we could actually disallow the latter entirely, even in unsafe code, right?

src/what-unsafe-does.md
```diff
@@ -18,6 +20,7 @@ language cares about is preventing the following things:
 
 * Dereferencing (using the `*` operator on) dangling or unaligned pointers (see below)
 * Breaking the [pointer aliasing rules][]
+* Violating the [const safety][] rules
```
RalfJung, line 23: Uh-oh... this list is synced with [this one](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) and changing it is a big deal.^^

src/const-safety.md
RalfJung: More importantly though, I feel the first sentence is not a great introduction here. I wrote this in the context of the const-eval repo, but this is different.

I'd start with something like "As part of the compilation process, the Rust compiler has to evaluate some of the code of a program. This is called CTFE (compile-time function evaluation). CTFE is used for: array lengths, const/static initializers, ...". Then say that CTFE can go wrong and just like at run-time, safe Rust protects programmers from some of the ways CTFE can fail. "Programmers only writing safe code do not have to care, but if you want to write unsafe code that is run by CTFE, then there are some extra concerns you need to be aware of."

Then the following text makes much more sense IMO.^^

src/const-safety.md
oli-obk: Yea, this is from before the `guaranteed_*` PR where I ended up agreeing that making it UB likely isn't worth it.

RalfJung:
So after the lang team design meeting that just happened, another alternative that emerged is to say that doing unconst things during CTFE is "simply" UB. And then we just have to describe what CTFE-UB is like, which we need to do anyway. I think that is actually my favorite plan so far. :D

Even if we adopt this (thus making most of the text here obsolete), some of the editing here could be incorporated into the const-eval version of this document.

JohnTitor:
Triage: It has been about a year since the last comment, I presume this has been waiting for @oli-obk to address the review comments, right?

RalfJung:
Also https://github.com/rust-lang/rfcs/pull/3016 landed in the mean time (though I am not sure if that will affect this PR).

src/what-unsafe-does.md
RalfJung: To be clear, what I am saying is that I think this PR should *not* change this list.

src/const-safety.md
```diff
@@ -0,0 +1,171 @@
...
+
+The modulo operation here is not supported by the miri engine because evaluating
+it requires knowing the actual integer address of `S`.
+
+The way miri handles this is by treating pointer and integer values separately.
+The most primitive kind of value in miri is a `Scalar`, and a scalar is *either*
+a pointer (`Scalar::Ptr`) *or* a bunch of bits representing an integer
+(`Scalar::Bits`), or uninitialized.  Every value of a variable of primitive type is stored as a
+`Scalar`.  In the code above, casting the pointer `&S` to `*const i32` and then
+to `usize` does not actually change the value -- we end up with a local variable
```
RalfJung, line 49: We don't allow ptr-to-int casts in `const` any more... should we adjust this story?

oli-obk: I'm going through the entire doc, too many things changed. I'm keeping everything in mind that was commented here, but the entire doc needs an overhaul

PR #231, opened 2020-07-28
fix a typo

lwintermelon:
it should be covariant, right?


nsmryan:
There is one additional use of the word 'variant'!
```
 The lifetime will be bounded, and your iterator will be variant over 'a and T. 
```
should be
```
 The lifetime will be bounded, and your iterator will be covariant over 'a and T. 
```

JohnTitor (approved):
Looks good! I'm going to merge this as-is, a follow-up PR is welcome :)

PR #232, opened 2020-08-03
Update casts.md

turtleslow:
Float to integer cast unsoundness was fixed in Rust 1.45.0 ([announcment](https://blog.rust-lang.org/2020/07/16/Rust-1.45.0.html#fixing-unsoundness-in-casts)).


JohnTitor (approved):
Looks good, thanks!

PR #240, opened 2020-10-17
Tweak GHA config

JohnTitor:


lzutao (approved):

.github/workflows/main.yml
```diff
@@ -1,12 +1,17 @@
 name: CI
-on: [push, pull_request]
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  push:
+    branches:
+      - master
```
ehuss, line 7: Can you explain more why these changes are made?

JohnTitor: This has little impact but it's mainly for not triggering CI twice on PRs when someone pushes commits to this repo, not their forked repo. Sometimes folks having write access make a branch to fix a typo or any minor change and it'll trigger two jobs when they submit a PR.
Also, set PRs trigger events explicitly, in case the default to trigger events is changed (reference: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request)

ehuss: But if we need to create branches for things like beta backports (unlikely, but possible), it won't run.  Also, I'm not sure it's a good idea to create random branches in this repo, since that impacts everyone.  And running CI a few extra times shouldn't have a negative impact, should it?

I personally would just leave the defaults, since they just add noise and complexity. It seems unlikely that the defaults will change, and if they do it seems possible that this would have the undesired effect of missing new events that should be running.

JohnTitor: > Also, I'm not sure it's a good idea to create random branches in this repo, since that impacts everyone.

I totally agree with this but editing something on GitHub creates a branch on the repo (for instance, https://github.com/rust-lang/rust/tree/LeSeulArtichaut-patch-1 is accidentally created).

> And running CI a few extra times shouldn't have a negative impact, should it?

Makes sense. I don't like running CI twice because I've seen it will cause spurious failure (for instance, 429 failure occurred on the rustc-dev-guide frequently and it was really noisy for us). But in this repo, CI isn't marked as required and it seems there's no spurious failure so running it multi times could be fine.

> But if we need to create branches for things like beta backports (unlikely, but possible), it won't run.

This should be caught indeed, yeah.

So, I'm fine to drop that commit for now, given the above points. Thanks!

JohnTitor:
Rebased and dropped the last commit, PTAL :)

JohnTitor:
Oh, we've encountered another failure but I'm not sure about this, maybe caused by the mdbook version?

ehuss:
Oh, the link checks are failing due to the update to mdbook.  To fix that, the environment variable `MDBOOK_OUTPUT__HTML__INPUT_404` can be set to an empty value.  It might be best to put that in the [linkcheck.sh](https://github.com/rust-lang/rust/blob/master/src/tools/linkchecker/linkcheck.sh) script directly. Would you be willing to make a PR to rust-lang/rust to add that? I can review/merge that.

JohnTitor:
Sure! Will do that soonish.

ehuss (approved):
Thanks!

PR #242, opened 2020-10-31
Improve formatting of "Transmutes" chapter

camelid:
* Use inline code style for `&` and `&mut`
* Add blank lines between list items to make source easier to read
* Use new-style mdBook link


JohnTitor (approved):
Thanks!

PR #245, opened 2020-11-02
Uppercase `repr(Rust)`

camelid:
There were two spots where it referred to `repr(rust)` when it is
usually referred to as `repr(Rust)` in accordance with `repr(C)`.
I updated those to use the uppercase form.


JohnTitor (approved):
Good catch, thanks!

PR #246, opened 2020-11-22
Clarify that any alignment is valid for ZSTs

nico-abram:
Fixes https://github.com/rust-lang/nomicon/issues/226


Gankra (approved):
thanks!

PR #251, opened 2020-12-30
Update vector code examples

lukas9393:
Hi, I tried to finish the chapter about Vec. Unfortunately this chapter contains old source code that is not buildable. I had difficulties to get the code to work. Finally I found the last chapter with the actual working code.

I took this and updated the code examples. With this code it should now be possible, step by step, to evolve to the final code.

### DIFF
--- a/src/vec-alloc.md
+++ b/src/vec-alloc.md
@@ -16,10 +16,6 @@ want to use `dangling` because there's no real allocation to talk about but
 So:
 
 ```rust,ignore
-#![feature(alloc, heap_api)]
-
-use std::mem;
-
 impl<T> Vec<T> {
     fn new() -> Self {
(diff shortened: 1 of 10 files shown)


Gankra:
Can you provide some insight into why you're using the unstable GlobalAlloc, and not the stable global alloc functions?

Gankra:
(this is hard to express properly because of the super overloaded terms -- these functions at the bottom: https://doc.rust-lang.org/alloc/alloc/index.html#functions)

lukas9393:
I have completely used the template from vec-final to update the tutorial.  I think the reason why the vec-final example does not use the stable API is that the stable api will be deprecated in the future.

Gankra (approved):
oh! i see, someone updated just the final part but not the code around it. ok, awesome!

PR #252, opened 2020-12-31
Remove outdated information about `jemalloc`

### LINKED ISSUE #229: Incorrect parenthetical statement that jemalloc is the default allocator

JohnTitor:
Closes #206
Fixes #229

This supersedes #206 since it mentions `malloc` but doesn't `HeapAlloc` (used on Windows).

### DIFF
--- a/src/vec-alloc.md
+++ b/src/vec-alloc.md
@@ -34,7 +34,8 @@ Without this assert, some of our early drafts will do some Very Bad Things.
 
 Next we need to figure out what to actually do when we *do* want space. For
 that, we'll need to use the rest of the heap APIs. These basically allow us to
-talk directly to Rust's allocator (jemalloc by default).
+talk directly to Rust's allocator (`malloc` on Unix platforms and `HeapAlloc`
+on Windows by default).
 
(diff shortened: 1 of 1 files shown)


Gankra (approved):
thanks!

JohnTitor:
Hey, @Gankra, the lang-docs team now maintains the language-related docs including nomicon, is it okay to add you the team so that I can ping you via the team mention?

Gankra:
Sure, although I have pretty limited bandwidth so will be a bit unreliable to say the least

JohnTitor:
Great, opened https://github.com/rust-lang/team/pull/508.

> although I have pretty limited bandwidth so will be a bit unreliable to say the least

It's totally fine, someone could review instead when you're unavailable (that's why we have the team).

PR #253, opened 2021-01-01
Simple Arc implementation (without Weak refs)

ThePuzzlemaker:
This is a simple implementation of `Arc`, without weak refs. If there's anything you feel that could be improved (or if I did anything wrong), please tell me. I'm open to recommendations :slightly_smiling_face:


src/arc.md
```diff
@@ -0,0 +1,13 @@
+# Implementing Arc
+
+In this section, we'll be implementing a simpler version of `std::sync::Arc`.
+Similarly to [the implementation of `Vec` we made earlier](vec.md), we won't be
+taking advantage of as many optimizations, intrinsics, or unstable code as the
+standard library may.
+
+This implementation is loosely based on the standard library's implementation
+(technically taken from `alloc::sync` in the latest nightly), but it will not
```
Havvy, line 9: The latest nightly will desync from this version eventually. You should probably give a date to this nightly and link to the file in the commit of the nightly you are talking about.

Havvy (changes requested):
Awesome that somebody is adding such a huge change; I just did a cursory look (this is not my area of knowledge), and I wasn't even going to comment, but one thing caught my eye so hard I had to.

src/arc-final.md
```diff
@@ -0,0 +1,80 @@
...
+    rc: AtomicUsize,
+    data: T,
+}
+
+impl<T> Arc<T> {
+    pub fn new(data: T) -> Arc<T> {
+        // We start the reference count at 1, as that first reference is the
+        // current pointer.
+        let boxed = Box::new(ArcInner {
+            rc: AtomicUsize::new(0),
```
Freax13, line 26: shouldn't this be 1 instead of 0?

ThePuzzlemaker: Thanks for catching that, I'm dumb :thinking:

src/arc.md
ThePuzzlemaker: I'll just make sure that nightly is in sync with stable right now and just use stable 1.49 as a base

src/arc-drop.md
```diff
@@ -0,0 +1,101 @@
...
+
+Basically, we need to:
+1. Get the `ArcInner` value of the `Arc`
+2. Decrement the reference count
+3. If there is only one reference remaining to the data, then:
+4. Atomically fence the data to prevent reordering of the use and deletion of
+   the data, then:
+5. Drop the inner data 
+
+We'll need to import the trait:
```
BlackHoleFox, line 16: The `Drop` trait is in [the standard library's prelude](https://doc.rust-lang.org/std/prelude/#prelude-contents), so it doesn't need explicitly imported, you can just write `impl Drop for Foo { ... }` anywhere.

ThePuzzlemaker: ah ok

Havvy (commented):
Saying it's from 1.49 sounds good to me.

Havvy (dismissed):

src/arc-drop.md
```diff
@@ -0,0 +1,92 @@
...
+previous value):
+```rust,ignore
+if self.inner().rc.fetch_sub(1, Ordering::Release) != 1 {
+    return;
+}
+```
+
+We then need to create an atomic fence to prevent reordering of the use of the
+data and deletion of the data. As described in [the standard library's
+implementation of `Arc`][3]:
```
Gankra, line 27: and change this to something like:

------

What atomic ordering should we use here? To know that, we need to consider what happens-before relationships we want to ensure (or alternatively, what dataraces we want to prevent). Drop is special because it's the one place where we mutate the Arc's payload (by dropping and freeing it). This is a potential read-write data race with all the other threads that have been happily reading the payload without any synchronization.

So we need to ensure all those non-atomic accesses have a proper happens-before relationship with us dropping and freeing the payload. To establish happens-before relationships with non-atomic accesses, we need (at least) [Acquire-Release semantics](https://doc.rust-lang.org/nomicon/atomics.html#acquire-release).

As a reminder, Acquires ensure non-atomic accesses after them on the same thread *stay* after them (they happen-before everything that comes after them) and Releases ensure non-atomic accesses before them on the same thread *stay* before them (everything before them happens-before them).

So we have many threads that look like this:

```
(A) non-atomic accesses to payload
(B) atomic decrement refcount
```

And a "final" thread that looks like this:

```
(C) non-atomic accesses to payload
(D) atomic decrement refcount
(E) non-atomic free/drop contents
```

And we want to ensure every thread agrees that everything happens-before E.

One thing that jumps out clearly is that the non-final threads all end with an atomic access (B), and we want to keep everything else (A) before it. That's exactly what a Release does! So it seems we'd like our atomic decrement to be a Release.

```
if self.inner().rc.fetch_sub(1, Ordering::Release) != 1 {
  return;
}
```


However this on its own doesn't work -- our final thread would *also* use a Release, and that means (E) would be allowed to happen-before (D)! To prevent this we need Release's partner, an Acquire! We *could* make (D) AcquireRelease (AcqRel), but this would penalize all the other threads performing (B). So instead, we will introduce a separate Acquire that only happens if we're the final thread. And since we've already loaded all the values we need, we can use a fence.

```
if self.inner().rc.fetch_sub(1, Ordering::Release) != 1 {
  return;
}
atomic::fence(Ordering::Acquire);
```

If this helps, you can think of this like a sort of implicit [RWLock](https://doc.rust-lang.org/std/sync/struct.RwLock.html): every Arc is a ReadGuard which allows unlimited read access until they go away and "Release the lock" (trapping all accesses on that thread before that point). The final thread then upgrades itself to a WriteGuard which "Acquires the lock" (creating a new critical section which strictly happens after everything else).

Gankra (changes requested):
Overall this is really great, but it definitely needs some revision.

src/arc-base.md
```diff
@@ -0,0 +1,104 @@
...
+            // `Box::into_raw` which is guaranteed to not be null.
+            ptr: NonNull::new(Box::into_raw(boxed)).unwrap(),
+            _marker: PhantomData,
+        }
+    }
+
+    fn inner(&self) -> &ArcInner<T> {
+        // This unsafety is okay because while this Arc is alive, we're
+        // guaranteed that the inner pointer is valid.
+        unsafe { self.ptr.as_ref() }
```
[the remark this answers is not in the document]
RalfJung: Not really, https://github.com/rust-lang/rust/issues/55005 is still open.

There has been a great proposal in https://github.com/rust-lang/unsafe-code-guidelines/issues/252 though that would fix this for `Arc`. However, it would mean references are not necessarily "dereferencable until the end" any more.

src/arc-drop.md
RalfJung: > However this on its own doesn't work -- our final thread would also use a Release, and that means (E) would be allowed to happen-before (D)!

This sounds like for X, Y we always have that one happens-before the other or vice versa... which is not the way happens-before actually works. Happens-before is a *partial order*, and some events simply are unordered. So I think it'd be better to phrase this accordingly, saying something like "[...] and that means (D) would not be forced to happen-before (E)". Except that's also wrong, program-order is included in happens-before. The actual issue is between (E) and (A), isn't it? We must make (A) happens-before (E), and that's why there needs to be an "acquire" in the "final" thread.

RalfJung: > As a reminder, Acquires ensure non-atomic accesses after them on the same thread stay after them (they happen-before everything that comes after them) and Releases ensure non-atomic accesses before them on the same thread stay before them (everything before them happens-before them).

The key thing with an acquire is that *the release it reads from* (and everything that comes before it) happens-before everything that comes after the acquire.  Each acquire is paired with a release, and this pair establishes a happens-before link *across threads*. Personally, I find this way of thinking about it easier than thinking about the release and the acquire separately. (Also see what I said above: to my knowledge, happens-before includes program-order, so "X happens-before everything that comes after it in the same thread" is true for all X.)

src/arc-base.md
```diff
@@ -0,0 +1,104 @@
...
+
+```rust,ignore
+unsafe impl<T: Sync + Send> Send for Arc<T> {}
+unsafe impl<T: Sync + Send> Sync for Arc<T> {}
+```
+
+We need to have the bound `T: Sync + Send` because if we did not provide those
+bounds, it would be possible to share values that are thread-unsafe across a
+thread boundary via an `Arc`, which could possibly cause data races or
+unsoundness.
```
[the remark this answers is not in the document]
ThePuzzlemaker: How does this sound?

> We need to have the bound `T: Sync + Send` because if we did not provide those bounds, it would be possible to share values that are thread-unsafe across a
thread boundary via an `Arc`, which could possibly cause data races or
unsoundness.
> 
> For example, if those bounds were not present, `Arc<Rc<u32>>` would be `Sync` or `Send`, meaning that you could clone the `Rc` out of the `Arc` to send it across a thread (without creating an entirely new `Rc`), which would create data races as `Rc` is not thread-safe.

src/arc-clone.md
```diff
@@ -0,0 +1,60 @@
...
+[2]: https://github.com/rust-lang/rust/blob/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/alloc/src/sync.rs#L1171-L1181
+
+We'll need to add another import to use `Ordering`:
+```rust,ignore
+use std::sync::atomic::Ordering;
+```
+
+It is possible that in some contrived programs (e.g. using `mem::forget`) that
+the reference count could overflow, but it's unreasonable that would happen in
+any reasonable program.
```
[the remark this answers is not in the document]
ThePuzzlemaker: I originally included this but decided it wasn't necessary. I didn't notice that whole section. Thanks!

src/arc-clone.md
```diff
@@ -0,0 +1,60 @@
...
+> 
+> As explained in the [Boost documentation][1]:
+> > Increasing the reference counter can always be done with
+> > memory_order_relaxed: New references to an object can only be formed from an
+> > existing reference, and passing an existing reference from one thread to
+> > another must already provide any required synchronization.
+> 
+> [1]: https://www.boost.org/doc/libs/1_55_0/doc/html/atomic/usage_examples.html
+[2]: https://github.com/rust-lang/rust/blob/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/alloc/src/sync.rs#L1171-L1181
+
```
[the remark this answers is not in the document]
ThePuzzlemaker: How does this sound? (I'm not exactly the best at atomics so I hope it's correct)
The links there are actually valid in the markdown file

---

But what ordering should we use here? We don't really have any code that will
need atomic synchronization when cloning, as we do not modify the internal value
while cloning. Thus, we can use a Relaxed ordering here, which implies no
happens-before relationship but is atomic. When `Drop`ping the Arc, however,
we'll need to atomically synchronize when decrementing the reference count. This
is described more in [the section on the `Drop` implementation for
`Arc`]() For more information on atomic relationships and Relaxed
ordering, see [the section on atomics]().

src/arc-base.md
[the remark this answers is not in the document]
ThePuzzlemaker: Is there anything I need to do related to this?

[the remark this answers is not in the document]
Gankra: Not really

Gankra:
(probably won't have a chance to review the new changes until monday)

src/arc-drop.md
```diff
@@ -0,0 +1,99 @@
...
+also return if the returned value from `fetch_sub` (the value of the reference
+count before decrementing it) is not equal to `1` (which happens when we are not
+the last reference to the data).
+```rust,ignore
+if inner.rc.fetch_sub(1, Ordering::???) != 1 {
+    return;
+}
+```
+
+TODO
```
Gankra, line 30: Drop this TODO and remove the ??? above for now -- let's land this as a good starting point we can improve later.

Gankra (changes requested):
Looks good!

src/arc-drop.md
ThePuzzlemaker: I was mostly waiting to implement your suggestions on this but got killed by the wall of text lol

Gankra (approved):

Gankra:
thanks a bunch!!

PR #254, opened 2021-01-18
Adds compiler error for example in Limits of Lifetimes

mdaverde:
Hello!

I wanted to see what the compiler error actually was for the given example under "Improperly reduced borrows" so I added it for future readers to see


JohnTitor (approved):
Nice improvement, thanks!

PR #255, opened 2021-01-21
Fix alloc link in exotic-sizes for local docs

cherryblossom000:
This changes `https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html#tymethod.alloc` to `../std/alloc/trait.GlobalAlloc.html#tymethod.alloc` so that it uses the local docs installation if the page is viewed with the local docs, instead of https://doc.rust-lang.org.


ehuss (approved):
Thanks!

PR #256, opened 2021-01-22
Update casts.md

edgeone89:
casting inf and -inf to integer


Havvy (commented):
These are just repeats of the previous two bullets before `NaN` but using specific values. Perhaps put "including `inf`" and ", including -inf" in those bullets instead?

edgeone89:
> These are just repeats of the previous two bullets before `NaN` but using specific values. Perhaps put "including `inf`" and ", including -inf" in those bullets instead?

maybe

JohnTitor:
@edgeone89 Ping, could you address @Havvy's comment? Thanks!

Issue #250, opened 2020-12-16
Described opaque type pattern has issues


skade:
It was recently pointed out that the current suggestion around representing opaque types in Rust has issues.

https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs

```rust
#[repr(C)] pub struct Bar { _private: [u8; 0] }
```

This type ends up being `Send`, `Sync` and `Unpin`, all properties which might not hold for the type hidden.

A type I came up with @dtolnay and @nvzqz to fix those issues is the following:

```rust
#[repr(C)]
pub struct opaque_example {
    // Required for FFI-safe 0-sized type.
    //
    // In the future, this should refer to an extern type.
    // See https://github.com/rust-lang/rust/issues/43467.
    _data: [u8; 0],

    // Required for !Send & !Sync & !Unpin.
    //
    // - `*mut u8` is !Send & !Sync. It must be in `PhantomData` to not
    //   affect alignment.
    //
    // - `PhantomPinned` is !Unpin. It must be in `PhantomData` because
    //   its memory representation is not considered FFI-safe.
    _marker:
        core::marker::PhantomData<(*mut u8, core::marker::PhantomPinned)>,
}
```

(An implementation can be found here: https://github.com/skade/ffi-opaque/)

For reference, this matches the current behaviour of RFC 1861 extern types: https://github.com/rust-lang/rust/pull/44295#issue-139057745

JohnTitor:
Makes sense, @skade do you have time to update the docs? If not I'm happy to work on it instead :)

PR #257, opened 2021-03-08
Fix opaque type representation.

skade:
Fix for #250.

Final pattern derived by @nvzqz, with help from @dtolnay.

### DIFF
--- a/src/ffi.md
+++ b/src/ffi.md
@@ -740,8 +740,18 @@ void bar(struct Bar *arg);
 To do this in Rust, let’s create our own opaque types:
 
 ```rust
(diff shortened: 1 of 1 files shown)


thomcc:
That pattern was actually fine though, wasn't it?

Since you never use `Bar` by value, you use it behind a raw pointer, which is thus !Send, !Sync, and Pinning wouldn't apply.

I think a bigger (and possibly more controversial) issue is not discouraging using these types by value/reference, which is likely UB under the current aliasing semantics (although it is a thing people do...)

As it is, I've seen people confused about the reasoning here, and IDK if this change makes things clearer.

PR #259, opened 2021-03-22
Add example of thinking about Send/Sync's soundness

danielzfranklin:
Add an example of thinking through whether it is sound to implement Send + Sync for a custom type that wraps a raw pointer.

I read the existing docs and was confused about whether I could implement Send + Sync for a type I wrote that wraps a c-style array. Kixiron, InfernoDeity, Talchas, and HeroicKatora on #black-magic helped me understand Send and Sync better. This example is based on the advice they gave me. I've made lots of changes, so any errors are probably mine.

### DIFF
--- a/src/send-and-sync.md
+++ b/src/send-and-sync.md
@@ -89,19 +89,31 @@ use std::ptr::NonNull;
 struct Carton<T>(NonNull<T>);
 
 impl<T> Carton<T> {
-    pub fn new(mut value: T) -> Self {
-        // Allocate enough memory on the heap to store one T
-        let ptr = unsafe { libc::calloc(1, size_of::<T>()) as *mut T };
+    pub fn new(value: T) -> Self {
+        // Allocate enough memory on the heap to store one T.
+        let memptr = &mut null_mut() as *mut *mut T;
+        unsafe {
+            let ret = libc::posix_memalign(
+                memptr as *mut *mut c_void,
+                align_of::<T>(),
+                size_of::<T>()
+            );
+            assert_eq!(ret, 0, "Failed to allocate or invalid alignment");
+        };
 
         // NonNull is just a wrapper that enforces that the pointer isn't null.
-        // Malloc returns null if it can't allocate.
-        let mut ptr = NonNull::new(ptr).expect("We assume malloc doesn't fail");
+        let mut ptr = unsafe {
+            // Safety: memptr is dereferenceable because we created it from a
+            // reference and have exclusive access.
+            NonNull::new(*memptr)
+                .expect("Guaranteed non-null if posix_memalign returns 0")
+        };
 
-        // Move value from the stack to the location we allocated on the heap
+        // Move value from the stack to the location we allocated on the heap.
         unsafe {
-            // Safety: The pointer returned by calloc is alligned, initialized,
-            // and dereferenceable, and we have exclusive access to the pointer.
-            *ptr.as_mut() = value;
+            // Safety: If non-null, posix_memalign gives us a ptr that is valid
+            // for writes and properly aligned.
+            ptr.as_ptr().write(value);
         }
 
         Self(ptr)
--- a/src/send-and-sync.md
+++ b/src/send-and-sync.md
@@ -93,7 +93,7 @@ struct Carton<T>(ptr::NonNull<T>);
 impl<T> Carton<T> {
     pub fn new(value: T) -> Self {
         // Allocate enough memory on the heap to store one T.
-        assert_ne!(size_of::<T>(), 0, "Zero-sized types are out of the scope of this example);
+        assert_ne!(size_of::<T>(), 0, "Zero-sized types are out of the scope of this example");
         let memptr: *mut c_void = ptr::null_mut();
         unsafe {
             let ret = libc::posix_memalign(
--- a/src/send-and-sync.md
+++ b/src/send-and-sync.md
@@ -108,7 +108,7 @@ impl<T> Carton<T> {
         let mut ptr = unsafe {
             // Safety: memptr is dereferenceable because we created it from a
             // reference and have exclusive access.
-            NonNull::new(memptr.cast::<T>())
+            ptr::NonNull::new(memptr.cast::<T>())
                 .expect("Guaranteed non-null if posix_memalign returns 0")
         };
 
--- a/src/send-and-sync.md
+++ b/src/send-and-sync.md
@@ -93,10 +93,11 @@ struct Carton<T>(NonNull<T>);
 impl<T> Carton<T> {
     pub fn new(value: T) -> Self {
         // Allocate enough memory on the heap to store one T.
-        let memptr = &mut null_mut() as *mut *mut T;
+        assert_ne!(size_of::<T>(), 0, "Zero-sized types are out of the scope of this example);
+        let memptr: *mut c_void = ptr::null_mut();
         unsafe {
             let ret = libc::posix_memalign(
-                memptr as *mut *mut c_void,
+                &mut memptr,
                 align_of::<T>(),
                 size_of::<T>()
             );
@@ -107,7 +108,7 @@ impl<T> Carton<T> {
         let mut ptr = unsafe {
             // Safety: memptr is dereferenceable because we created it from a
             // reference and have exclusive access.
-            NonNull::new(*memptr)
+            NonNull::new(memptr.cast::<T>())
                 .expect("Guaranteed non-null if posix_memalign returns 0")
         };
 
--- a/src/send-and-sync.md
+++ b/src/send-and-sync.md
@@ -126,8 +126,10 @@ to access it. [`Box`][box-doc] implements [`Deref`][deref-doc] and
 [`DerefMut`][deref-mut-doc] so that you can access the inner value. Let's do
 that.
 
-```rust,ignore
+```rust
 use std::ops::{Deref, DerefMut};
+
+# struct Carton<T>(std::ptr::NonNull<T>);
 
 impl<T> Deref for Carton<T> {
     type Target = T;
@@ -165,7 +167,8 @@ safely be Send unless it shares mutable state with something else without
 enforcing exclusive access to it. Each `Carton` has a unique pointer, so
 we're good.
 
-```rust,ignore
+```rust
+# struct Carton<T>(std::ptr::NonNull<T>);
 // Safety: No one besides us has the raw pointer, so we can safely transfer the
 // Carton to another thread if T can be safely transferred.
 unsafe impl<T> Send for Carton<T> where T: Send {}
@@ -178,7 +181,8 @@ write to the pointer, and the borrow checker enforces that mutable
 references must be exclusive, there are no soundness issues making `Carton`
 sync either.
 
-```rust,ignore
+```rust
+# struct Carton<T>(std::ptr::NonNull<T>);
 // Safety: Our implementation of DerefMut requires writers to mutably borrow the
 // Carton, so the borrow checker will only let us have references to the Carton
 // on multiple threads if no one has a mutable reference to the Carton. This
--- a/src/send-and-sync.md
+++ b/src/send-and-sync.md
@@ -85,25 +85,27 @@ We start by writing code to take a value allocated on the stack and transfer it
 to the heap.
 
 ```rust
-# mod libc {
-#     pub use ::std::os::raw::{c_int, c_void, size_t};
-#     extern "C" { pub fn posix_memalign(memptr: *mut *mut c_void, align: size_t, size: size_t) -> c_int; }
+# pub mod libc {
+#    pub use ::std::os::raw::{c_int, c_void};
+#    #[allow(non_camel_case_types)]
+#    pub type size_t = usize;
+#    extern "C" { pub fn posix_memalign(memptr: *mut *mut c_void, align: size_t, size: size_t) -> c_int; }
 # }
 use std::{
     mem::{align_of, size_of},
     ptr,
 };
-use libc::c_void;
+
 struct Carton<T>(ptr::NonNull<T>);
 
 impl<T> Carton<T> {
     pub fn new(value: T) -> Self {
         // Allocate enough memory on the heap to store one T.
         assert_ne!(size_of::<T>(), 0, "Zero-sized types are out of the scope of this example");
-        let memptr: *mut c_void = ptr::null_mut();
+        let mut memptr = ptr::null_mut() as *mut T;
         unsafe {
             let ret = libc::posix_memalign(
-                &mut memptr,
+                (&mut memptr).cast(),
                 align_of::<T>(),
                 size_of::<T>()
             );
@@ -218,7 +220,12 @@ we need to know `free` can be called on a pointer that was yielded by an
 allocation done on another thread. We can check this is true in the docs for
 [`libc::free`][libc-free-docs].
 
-```rust,ignore
+```rust
+# struct Carton<T>(std::ptr::NonNull<T>);
+# mod libc {
+#     pub use ::std::os::raw::c_void;
+#     extern "C" { pub fn free(p: *mut c_void); }
+# }
 impl<T> Drop for Carton<T> {
     fn drop(&mut self) {
         unsafe {
--- a/src/send-and-sync.md
+++ b/src/send-and-sync.md
@@ -86,9 +86,9 @@ to the heap.
 
 ```rust,ignore
 use std::mem::size_of;
-use std::ptr::NonNull;
+use std::ptr;
 
-struct Carton<T>(NonNull<T>);
+struct Carton<T>(ptr::NonNull<T>);
 
 impl<T> Carton<T> {
     pub fn new(value: T) -> Self {
--- a/src/send-and-sync.md
+++ b/src/send-and-sync.md
@@ -209,8 +209,8 @@ unsafe impl<T> Send for Carton<T> where Box<T>: Send {}
 Right now Carton<T> has a memory leak, as it never frees the memory it allocates.
 Once we fix that we have a new requirement we have to ensure we meet to be Send:
 we need to know `free` can be called on a pointer that was yielded by an
-allocation done on another thread. This is the case for `libc::free`, so we can
-still be Send.
+allocation done on another thread. We can check this is true in the docs for
+[`libc::free`][libc-free-docs].
 
 ```rust,ignore
 impl<T> Drop for Carton<T> {
@@ -227,7 +227,7 @@ A nice example where this does not happen is with a MutexGuard: notice how
 [uses libraries][mutex-guard-not-send-comment] that require you to ensure you
 don't try to free a lock that you acquired in a different thread. If you were
 able to Send a MutexGuard to another thread the destructor would run in the
-thread you sent it to, violating the requirement. MutexGuard can still be sync
+thread you sent it to, violating the requirement. MutexGuard can still be Sync
 because all you can send to another thread is an `&MutexGuard` and dropping a
 reference does nothing.
 
@@ -241,3 +241,4 @@ only to data races?
 [deref-mut-doc]: https://doc.rust-lang.org/core/ops/trait.DerefMut.html
 [mutex-guard-not-send-docs-rs]: https://doc.rust-lang.org/std/sync/struct.MutexGuard.html#impl-Send
 [mutex-guard-not-send-comment]: https://github.com/rust-lang/rust/issues/23465#issuecomment-82730326
+[libc-free-docs]: https://linux.die.net/man/3/free
--- a/src/send-and-sync.md
+++ b/src/send-and-sync.md
@@ -205,11 +205,7 @@ TODO: better explain what can or can't be Send or Sync. Sufficient to appeal
 only to data races?
 
 [unsafe traits]: safe-unsafe-meaning.html
-
 [box-doc]: https://doc.rust-lang.org/std/boxed/struct.Box.html
-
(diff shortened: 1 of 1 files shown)


danielzfranklin:
I think my example would be improved by an example of a change that would make Carton Send but not Sync and neither Send nor Sync, but I can't think of any that don't violate the safety invarients of casting pointers.

src/send-and-sync.md
```diff
@@ -74,7 +74,114 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+```
+
+This isn't very useful, because once our users give us a value they have no way
+to access it. [`Box`][box-doc] implements [`Deref`][deref-doc] and
+[`DerefMut`][deref-mut-doc] so that you can access the inner value. Let's do
+that.
+
+```rust,ignore
+use std::ops::{Deref, DerefMut};
+
```
JohnTitor, line 119:
```suggestion
```rust
use std::ops::{Deref, DerefMut};

# struct Carton<T>(std::ptr::NonNull<T>);
```

src/send-and-sync.md
```diff
@@ -74,7 +74,114 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+    }
+}
+```
+
+Finally, lets think about whether our `Carton` is Send and Sync. Something can
+safely be Send unless it shares mutable state with something else without
+enforcing exclusive access to it. Each `Carton` has a unique pointer, so
+we're good.
+
+```rust,ignore
```
JohnTitor, line 156:
```suggestion
```rust
# struct Carton<T>(std::ptr::NonNull<T>);
```

src/send-and-sync.md
```diff
@@ -74,7 +74,114 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+```
+
+What about Sync? For `Carton` to be Sync we have to enforce that you can't
+write to something stored in a `&Carton` while that same something could be read
+or written to from another `&Carton`. Since you need an `&mut Carton` to
+write to the pointer, and the borrow checker enforces that mutable
+references must be exclusive, there are no soundness issues making `Carton`
+sync either.
+
+```rust,ignore
```
JohnTitor, line 169:
```suggestion
```rust
# struct Carton<T>(std::ptr::NonNull<T>);
```

src/send-and-sync.md
```diff
@@ -74,7 +74,114 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
 Similarly, most iterators into these collections are Send and Sync because they
 largely behave like an `&` or `&mut` into the collection.
 
+[`Box`][box-doc] is implemented as it's own special intrinsic type by the
+compiler for [various reasons][box-is-special], but we can implement something
+with similar-ish behaviour ourselves to see an example of when it is sound to
```
JohnTitor, line 79:
```suggestion
with similar-ish behavior ourselves to see an example of when it is sound to
```

We usually prefer American-spelling in the docs.

src/send-and-sync.md
```diff
@@ -74,7 +74,114 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
 
 [unsafe traits]: safe-unsafe-meaning.html
+
+[box-doc]: https://doc.rust-lang.org/std/boxed/struct.Box.html
+
+[box-is-special]: https://manishearth.github.io/blog/2017/01/10/rust-tidbits-box-is-special/
+
+[deref-doc]: https://doc.rust-lang.org/core/ops/trait.Deref.html
+
+[deref-mut-doc]: https://doc.rust-lang.org/core/ops/trait.DerefMut.html
```
JohnTitor, line 187:
```suggestion
[box-doc]: https://doc.rust-lang.org/std/boxed/struct.Box.html
[box-is-special]: https://manishearth.github.io/blog/2017/01/10/rust-tidbits-box-is-special/
[deref-doc]: https://doc.rust-lang.org/core/ops/trait.Deref.html
[deref-mut-doc]: https://doc.rust-lang.org/core/ops/trait.DerefMut.html
```

src/send-and-sync.md
```diff
@@ -74,7 +74,114 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+        // Allocate enough memory on the heap to store one T
+        let ptr = unsafe { libc::calloc(1, size_of::<T>()) as *mut T };
+
+        // NonNull is just a wrapper that enforces that the pointer isn't null.
+        // Malloc returns null if it can't allocate.
+        let mut ptr = NonNull::new(ptr).expect("We assume malloc doesn't fail");
+
+        // Move value from the stack to the location we allocated on the heap
+        unsafe {
+            // Safety: The pointer returned by calloc is alligned, initialized,
```
danielhenrymantilla, line 102: I don't believe `calloc` always yields well-aligned pointers. AFAIK, `calloc` can yield addresses that are not aligned to `4 * size_of::<usize>`, for instance. That is, on a 32-bit machine, you may get pointers that are not `0x10`-aligned.

I thus suggest you put an assertion in there that `align_of::<T>` is small enough

src/send-and-sync.md
```diff
@@ -74,7 +74,114 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+            // and dereferenceable, and we have exclusive access to the pointer.
+            *ptr.as_mut() = value;
```
danielhenrymantilla, line 104:
```suggestion
            ptr.as_ptr().write(value);
```
Otherwise this assumes the pointee was already initialized with a valid bit-pattern for `T` (consider the case where `T` itself must be `NonNull`), and if it had drop glue, it will drop it!

src/send-and-sync.md
```diff
@@ -74,7 +74,114 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
 Similarly, most iterators into these collections are Send and Sync because they
 largely behave like an `&` or `&mut` into the collection.
 
+[`Box`][box-doc] is implemented as it's own special intrinsic type by the
```
JohnTitor, line 77: Could you add an h2 headering here? I think this should be separated.

src/send-and-sync.md
```diff
@@ -74,7 +74,114 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+impl<T> Carton<T> {
+    pub fn new(mut value: T) -> Self {
+        // Allocate enough memory on the heap to store one T
+        let ptr = unsafe { libc::calloc(1, size_of::<T>()) as *mut T };
+
+        // NonNull is just a wrapper that enforces that the pointer isn't null.
+        // Malloc returns null if it can't allocate.
+        let mut ptr = NonNull::new(ptr).expect("We assume malloc doesn't fail");
+
+        // Move value from the stack to the location we allocated on the heap
```
JohnTitor, line 100:
```suggestion
        // Move value from the stack to the location we allocated on the heap.
```

src/send-and-sync.md
```diff
@@ -74,7 +74,114 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+
+Finally, lets think about whether our `Carton` is Send and Sync. Something can
+safely be Send unless it shares mutable state with something else without
+enforcing exclusive access to it. Each `Carton` has a unique pointer, so
+we're good.
+
+```rust,ignore
+// Safety: No one besides us has the raw pointer, so we can safely transfer the
+// Carton to another thread.
+unsafe impl<T> Send for Carton<T> {}
```
danielhenrymantilla, line 159: What happens if the inner `T` is not `Send`? With `Carton<T>` you'd be able to send such a value to another thread.
```suggestion
unsafe impl<T : Send> Send for Carton<T> {}
```

src/send-and-sync.md
```diff
@@ -74,7 +74,114 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+or written to from another `&Carton`. Since you need an `&mut Carton` to
+write to the pointer, and the borrow checker enforces that mutable
+references must be exclusive, there are no soundness issues making `Carton`
+sync either.
+
+```rust,ignore
+// Safety: Our implementation of DerefMut requires writers to mutably borrow the
+// Carton, so the borrow checker will only let us have references to the Carton
+// on multiple threads if no one has a mutable reference to the Carton.
+unsafe impl<T> Sync for Carton<T> {}
```
danielhenrymantilla, line 173: Similarly,
```suggestion
unsafe impl<T : Sync> Sync for Carton<T> {}
```

JohnTitor (changes requested):
Thanks for writing up it! As a quick review, left some suggestions.

danielhenrymantilla (changes requested):
Careful, there are 4 soundness issues in your `Carton`:
  - `T` may require an alignment bigger than the one `calloc` delivers;
  - `mem::zeroed::<T>` may not be a _valid_ instance of type `T`, or it may even just not be a _safe_ instance, so it should not be `&mut`-overwritten;
  - Missing `Send` bound on the typre parameter when implementing `Send`
  - Ditto for `Sync`

If this is gonna be a tutorial, then I recommend you feature a very nice trick to avoid these `Send` / `Sync` accidents: when writing a type that is, _conceptually_, like a stdlib type, then write an impl based on that:
```diff
  impl<T> Send for Carton<T>
  where // when
+     Box<T> : Send,
  {}
```

danielzfranklin:
Thank you everyone. I appreciate you all taking the time to point out my mistakes.

I have one question. I used rust,ignore because I got a build error that libc wasn't available and I'd otherwise a lot of hidden duplicate lines to make each example compile on its own. I saw a lot of other examples do the same. Is there a different solution I should do?

JohnTitor:
> I have one question. I used rust,ignore because I got a build error that libc wasn't available and I'd otherwise a lot of hidden duplicate lines to make each example compile on its own. I saw a lot of other examples do the same. Is there a different solution I should do?

I'd like to avoid using `rust,ignore` so that we can notice if it's stalled. And note that I don't put suggestions for the snippet that uses `libc` but for the snippets that don't have any external deps. Indeed, we have a bunch of `rust,ignore` snippets but I'm planning to re-evaluate and clean-up it once I get some time to it.

src/send-and-sync.md
danielzfranklin: Thanks! I fixed this by using `posix_memalign` in c3ba178

src/send-and-sync.md
danielzfranklin: Fixed in c3ba178

src/send-and-sync.md
danielzfranklin: Fixed in 05a2df5

src/send-and-sync.md
danielzfranklin: Fixed in 05a2df5

src/send-and-sync.md
danielzfranklin: Fixed in 29f7139

src/send-and-sync.md
danielzfranklin: Fixed in 29f7139

src/send-and-sync.md
danielzfranklin: Fixed in 29f7139

src/send-and-sync.md
danielzfranklin: Fixed in c3ba178

src/send-and-sync.md
danielzfranklin: Fixed in e542c32

src/send-and-sync.md
```diff
@@ -74,7 +74,142 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+write to the pointer, and the borrow checker enforces that mutable
+references must be exclusive, there are no soundness issues making `Carton`
+sync either.
+
+```rust
+# struct Carton<T>(std::ptr::NonNull<T>);
+// Safety: Our implementation of DerefMut requires writers to mutably borrow the
+// Carton, so the borrow checker will only let us have references to the Carton
+// on multiple threads if no one has a mutable reference to the Carton. This
+// means we are Sync if T is Sync.
```
danielhenrymantilla, line 191:
```suggestion
// Safety: Since there exists a public way to go from a `&Carton<T>` to a `&T`
// in an unsynchronized fashion (such as `Deref`), then `Carton<T>` can't be
// `Sync` if `T` isn't.
// Conversely, `Carton` itself does not use any interior mutability whatsoever:
// all the mutations are performed through an exclusive reference (`&mut`). This
// means it suffices that `T` be `Sync` for `Carton<T>` to be `Sync`:
```

src/send-and-sync.md
```diff
@@ -74,7 +74,142 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+Finally, lets think about whether our `Carton` is Send and Sync. Something can
+safely be Send unless it shares mutable state with something else without
+enforcing exclusive access to it. Each `Carton` has a unique pointer, so
+we're good.
+
+```rust
+# struct Carton<T>(std::ptr::NonNull<T>);
+// Safety: No one besides us has the raw pointer, so we can safely transfer the
+// Carton to another thread if T can be safely transferred.
+unsafe impl<T> Send for Carton<T> where T: Send {}
```
danielhenrymantilla, line 176: Right now `Carton<T>` does not implement `Drop`, so there is a memory leak. Once it does (with a call to `libc::free(self.0.as_ptr().cast());`, part of the requirements of `Send` would be that `free` itself can be called on a pointer that yielded by an allocation done on another thread. This is the case of `free`, so we may forget to mention it, but it's a requirement nonetheless.

A nice example where this does not happen is with a [`MutexGuard`](https://doc.rust-lang.org/std/sync/struct.MutexGuard.html): notice how [it is not `Send`](https://doc.rust-lang.org/std/sync/struct.MutexGuard.html#impl-Send). Indeed, the Rust implementation of mutexes make use of functions (`pthread_ that explicitly mention the handles they yield cannot be used from another thread, even if the access is unique.

  - See also [this comment](https://github.com/rust-lang/rust/issues/23465#issuecomment-82730326)

src/send-and-sync.md
```diff
@@ -74,7 +74,140 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+compiler for [various reasons][box-is-special], but we can implement something
+with similar-ish behavior ourselves to see an example of when it is sound to
+implement Send and Sync. Let's call it a `Carton`.
+
+We start by writing code to take a value allocated on the stack and transfer it
+to the heap.
+
+```rust,ignore
+use std::mem::size_of;
+use std::ptr::NonNull;
```
danielhenrymantilla, line 89:
```suggestion
use std::ptr;
```

It is customary to use the functions within the `ptr` module with that `ptr` prefix: they read much better: consider `NonNull<usize>`, where it is not clear what is not non-zero, and `ptr::NonNull< usize >`. Similarly, `ptr::null_mut()` reads way better than `null_mut()` 🙂

danielhenrymantilla (commented):
Don't forget to `impl Drop`!

src/send-and-sync.md
danielzfranklin: Fixed

src/send-and-sync.md
danielzfranklin: That's a really good explanation. I tweaked it a little and added it to the page.

src/send-and-sync.md
```diff
@@ -74,7 +74,171 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+```rust,ignore
+use std::mem::size_of;
+use std::ptr;
+
+struct Carton<T>(ptr::NonNull<T>);
+
+impl<T> Carton<T> {
+    pub fn new(value: T) -> Self {
+        // Allocate enough memory on the heap to store one T.
+        assert_ne!(size_of::<T>(), 0, "Zero-sized types are out of the scope of this example);
```
danielhenrymantilla, line 96:
```suggestion
        assert_ne!(size_of::<T>(), 0, "Zero-sized types are out of the scope of this example");
```

Whoops, my suggestion had a typo in there

src/send-and-sync.md
```diff
@@ -74,7 +74,171 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+```
+
+A nice example where this does not happen is with a MutexGuard: notice how
+[it is not Send][mutex-guard-not-send-docs-rs]. The implementation of MutexGuard
+[uses libraries][mutex-guard-not-send-comment] that require you to ensure you
+don't try to free a lock that you acquired in a different thread. If you were
+able to Send a MutexGuard to another thread the destructor would run in the
+thread you sent it to, violating the requirement. MutexGuard can still be Sync
+because all you can send to another thread is an `&MutexGuard` and dropping a
+reference does nothing.
```
danielhenrymantilla, line 232: 👌

danielzfranklin:
I believe all the issues are addressed.

src/send-and-sync.md
```diff
@@ -74,7 +74,171 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+with similar-ish behavior ourselves to see an example of when it is sound to
+implement Send and Sync. Let's call it a `Carton`.
+
+We start by writing code to take a value allocated on the stack and transfer it
+to the heap.
+
+```rust,ignore
+use std::mem::size_of;
+use std::ptr;
+
```
danielhenrymantilla, line 90: Let's try to have this code be `cargo check`ed by mocking `libc`:
```suggestion
```rust
# mod libc {
#     pub use ::std::os::raw::{c_int, c_void, size_t};
#     extern "C" { pub fn posix_memalign(memptr: *mut *mut c_void, align: size_t, size: size_t) -> c_int; }
# }
use std::{
    mem::{align_of, size_of},
    ptr,
};
use libc::c_void;

```

src/send-and-sync.md
```diff
@@ -74,7 +74,171 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+                size_of::<T>()
+            );
+            assert_eq!(ret, 0, "Failed to allocate or invalid alignment");
+        };
+
+        // NonNull is just a wrapper that enforces that the pointer isn't null.
+        let mut ptr = unsafe {
+            // Safety: memptr is dereferenceable because we created it from a
+            // reference and have exclusive access.
+            NonNull::new(memptr.cast::<T>())
```
danielhenrymantilla, line 111:
```suggestion
            ptr::NonNull::new(memptr.cast::<T>())
```

danielhenrymantilla (approved):
The main constructor example wasn't "`cargo check`-ed", and I think it shouldn't take too much effort to fix that (just mock `libc` by using a module, as suggested). Other than that, the code snippets are sound, and the comments seem appropriate: LGTM

danielzfranklin:
I think this is ready @JohnTitor . Thank you everyone for helping me fix this!

src/send-and-sync.md
```diff
@@ -74,7 +74,184 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+This isn't very useful, because once our users give us a value they have no way
+to access it. [`Box`][box-doc] implements [`Deref`][deref-doc] and
+[`DerefMut`][deref-mut-doc] so that you can access the inner value. Let's do
+that.
+
+```rust
+use std::ops::{Deref, DerefMut};
+
+# struct Carton<T>(std::ptr::NonNull<T>);
+
```
JohnTitor, line 144:
```suggestion
#
```

Let's hide this line so that we don't show two blank lines.

src/send-and-sync.md
```diff
@@ -74,7 +74,184 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+            self.0.as_ref()
+        }
+    }
+}
+
+impl<T> DerefMut for Carton<T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        unsafe {
+            // Safety: The pointer is aligned, initialized, and dereferenceable
+            //   by the logic in [`Self::new]. We require writers to mutably
```
JohnTitor, line 165:
```suggestion
            //   by the logic in [`Self::new`]. We require writers to mutably
```

src/send-and-sync.md
```diff
@@ -74,7 +74,184 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+            //   elided to the lifetime of the input. This means the borrow
+            //   checker will enforce that no one else can access the contents
+            //   of the Carton until the mutable reference returned is dropped.
+            self.0.as_mut()
+        }
+    }
+}
+```
+
+Finally, lets think about whether our `Carton` is Send and Sync. Something can
```
JohnTitor, line 176: This is tooo nit but...

```suggestion
Finally, let's think about whether our `Carton` is Send and Sync. Something can
```

src/send-and-sync.md
```diff
@@ -74,7 +74,184 @@ of their pervasive use of raw pointers to manage allocations and complex ownersh
...
+standard library types we can assert that we have the same requirements.
+For example, the following code asserts that a Carton is Send if the same
+sort of Box would be Send, which in this case is the same as saying T is Send.
+
+```rust
+# struct Carton<T>(std::ptr::NonNull<T>);
+unsafe impl<T> Send for Carton<T> where Box<T>: Send {}
+```
+
+Right now Carton<T> has a memory leak, as it never frees the memory it allocates.
```
JohnTitor, line 217:
```suggestion
Right now `Carton<T>` has a memory leak, as it never frees the memory it allocates.
```

Otherwise it isn't rendered correctly.

JohnTitor (approved):
Thanks! Just read and it looks great to me with some nits.

danielzfranklin:
Thanks for catching those issues. All fixed.

JohnTitor:
Thanks again for the great work!

danielzfranklin:
Thank you for holding my hand so much! I've learned a lot.

PR #264, opened 2021-04-30
Update link to c++ atomic ordering docs

mlodato517:
I'm not sure if this is actually desirable. For one thing, the link I used isn't
locked to c++20 exactly. But it is the same page the [std lib](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html) links to
so that may be good for something.

I'm also not sure if other references to `C++11` should be updated or removed.
Basically, I feel like _something_ should change here but I'm not really sure what :-\


JohnTitor (approved):
Thanks!

> I'm also not sure if other references to C++11 should be updated or removed.

Feel free to open another PR if you like.

PR #265, opened 2021-05-02
Use pointer 'add' instead of 'offset'

blkerby:
This a clean-up to use the [`add`](https://doc.rust-lang.org/std/primitive.pointer.html#method.add) method on raw pointers where applicable, i.e., where [`offset`](https://doc.rust-lang.org/std/primitive.pointer.html#method.offset) was being used with a cast on the index from `usize` to `isize`. This implements a [suggestion](https://github.com/rust-lang/nomicon/pull/223#discussion_r615133425) from Amanieu.

### DIFF
--- a/src/exception-safety.md
+++ b/src/exception-safety.md
@@ -48,7 +48,7 @@ impl<T: Clone> Vec<T> {
             self.set_len(self.len() + to_push.len());
 
             for (i, x) in to_push.iter().enumerate() {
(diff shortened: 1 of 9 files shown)


JohnTitor:
cc/r? @Amanieu

Amanieu:
LGTM

@bors r+

Amanieu (approved):

JohnTitor (approved):
Just grepped and all the uses should be addressed, thanks!

PR #268, opened 2021-05-11
Clarify some of the language around marking traits safe/unsafe.

yaymukund:
> `Send` and `Sync` are marked unsafe because thread safety is a *fundamental property* that unsafe code can't
possibly hope to defend against in the way it could defend against a buggy `Ord` implementation.

When I initially read this, I thought it was strange you'd *want* to mark `Send` and `Sync` as `unsafe` if "unsafe code can't possibly hope to defend against [thread safety]".

Here, I've attempted to clarify.

### DIFF
--- a/src/safe-unsafe-meaning.md
+++ b/src/safe-unsafe-meaning.md
@@ -121,16 +121,19 @@ trait expects. In this situation, the Unsafe Rust in the internals of
 implementation is correct. If it isn't, it's the fault of the unsafe trait
 implementation, which is consistent with Rust's safety guarantees.
 
-The decision of whether to mark a trait `unsafe` is an API design choice.
-Rust has traditionally avoided doing this because it makes Unsafe
-Rust pervasive, which isn't desirable. `Send` and `Sync` are marked unsafe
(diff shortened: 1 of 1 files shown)


src/safe-unsafe-meaning.md
```diff
@@ -121,16 +121,19 @@ trait expects. In this situation, the Unsafe Rust in the internals of
...
+responsibility to the implementor. Rust has traditionally avoided marking
+traits `unsafe` because it makes Unsafe Rust pervasive, which isn't desirable.
+
+`Send` and `Sync` are marked unsafe because thread safety is a *fundamental
+property* that unsafe code can't possibly hope to defend against in the way it
+could defend against a buggy `Ord` implementation. Similarly, `GlobalAllocator`
+is keeping accounts of all the memory in the program and other things like
+`Box` or `Vec` build on top of it. If it does something weird (giving the same
+chunk of memory to another request when it is still in use), there's no chance
+to detect that and do anything about it.
```
yaymukund, line 136: This is all unchanged, I just split it out to a new paragraph for sake of readability.

PR #272, opened 2021-06-09
Clarify the conditions on the aliasing section

### LINKED ISSUE #148: In chapter 3.2, `Why Aliasing Matters` section, is the optimize example wrong?
In chapter 3.2([Aliasing](https://github.com/rust-lang-nursery/nomicon/blob/master/src/aliasing.md)), `Why Aliasing Matters` section, is the optimize example wrong?
[](url)

```rust
fn compute(input: &u32, output: &mut u32) {
    let cached_input = *input; // keep *input in a register
    if cached_input > 10 {
        *output = 2;  // x > 10 implies x > 5, so double and exit immediately
    } else if cached_input > 5 {
        *output *= 2;
    }
}
```

should be:
```rust
fn compute(input: &u32, output: &mut u32) {
    let cached_input = *input; // keep *input in a register
    if cached_input > 10 {
        *output *= 2;  // x > 10 implies x > 5, so double and exit immediately
    } else if cached_input > 5 {
        *output *= 2;
    }
}
```

or

```rust
fn compute(input: &u32, output: &mut u32) {
    let cached_input = *input; // keep *input in a register
    if cached_input > 10 {
        *output = 1;  // x > 10 implies x > 5, so double and exit immediately
    } else if cached_input > 5 {
        *output *= 2;
    }
}
```

JohnTitor:
Fixes #148

### DIFF
--- a/src/aliasing.md
+++ b/src/aliasing.md
@@ -31,6 +31,7 @@ fn compute(input: &u32, output: &mut u32) {
     if *input > 5 {
         *output *= 2;
     }
+    // remember that `output` will be `2` if `input > 10`
 }
 ```
 
@@ -38,9 +39,12 @@ We would *like* to be able to optimize it to the following function:
 
 ```rust
 fn compute(input: &u32, output: &mut u32) {
-    let cached_input = *input; // keep *input in a register
+    let cached_input = *input; // keep `*input` in a register
     if cached_input > 10 {
-        *output = 2;  // x > 10 implies x > 5, so double and exit immediately
+        // If the input is greater than 10, the previous code would set the output to 1 and then double it,
+        // resulting in an output of 2 (because `>10` implies `>5`).
+        // Here, we avoid the double assignment and just set it directly to 2.
+        *output = 2;
     } else if cached_input > 5 {
         *output *= 2;
     }


src/aliasing.md
```diff
@@ -31,16 +31,19 @@ fn compute(input: &u32, output: &mut u32) {
...
 We would *like* to be able to optimize it to the following function:
 
 ```rust
 fn compute(input: &u32, output: &mut u32) {
-    let cached_input = *input; // keep *input in a register
+    let cached_input = *input; // keep `*input` in a register
     if cached_input > 10 {
-        *output = 2;  // x > 10 implies x > 5, so double and exit immediately
+        // x > 10 implies x > 5, so remove the first if condition,
+        // set it to `2`, and exit immediately
```
ehuss, line 45: I'm still a little confused by this comment.  What is the "first if condition"?  The way I read it, the first `if` condition is `*input > 10`, which is still being checked here.  Can you clarify?

Also, this is using a new variable named `x` that isn't defined anywhere. It is somewhat obvious to me what it means, but it seems a little imprecise.

Perhaps something like this?

> If the input is greater than 10, the previous code would set the output to 1 and then double it, resulting in an output of 2 (because `>10` implies `>5`). Here, we avoid the double assignment and just set it directly to 2.

JohnTitor: Looks better, fixed.

ehuss (approved):
👍

PR #273, opened 2021-06-09
Mention "extern types" on the opaque structs section

### LINKED ISSUE #87: Update advice about opaque structs
Surely [this section](https://doc.rust-lang.org/nightly/nomicon/ffi.html#representing-opaque-structs) of the nomicon should now suggest `extern type` [given they're in nightly now](https://github.com/rust-lang/rust/issues/43467)? :-)

JohnTitor:
Fixes #87

### DIFF
--- a/src/ffi.md
+++ b/src/ffi.md
@@ -704,15 +704,8 @@ for more information.
 
 # Representing opaque structs
 
-Sometimes, a C library wants to provide a pointer to something, but not let you
-know the internal details of the thing it wants. The simplest way is to use "extern types".
-But it's currently (as of June 2021) unstable and has some unresolved questions,
-see the [RFC page][extern-type-rfc] and the [tracking issue][extern-type-issue] for more details.
-
-[extern-type-issue]: https://github.com/rust-lang/rust/issues/43467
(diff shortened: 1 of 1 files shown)


src/ffi.md
```diff
@@ -705,8 +705,14 @@ for more information.
 # Representing opaque structs
 
 Sometimes, a C library wants to provide a pointer to something, but not let you
-know the internal details of the thing it wants. The simplest way is to use a
-`void *` argument:
+know the internal details of the thing it wants. The simplest way is to use "extern types".
+But it's currently (as of June 2021) unstable and has some unresolved questions,
+see the [RFC page][extern-type-rfc] and the [tracking issue][extern-type-issue] for more details.
```
ehuss, line 710: Hmm, instead of starting off with talking about something that is unstable and likely something people can't use, what do you think about moving into a side note?  That is, take what you have written here, and move it down to the bottom of the page and make it a blockquote.

JohnTitor: Sounds good! Addressed via 538345a.

ehuss (approved):
👍

PR #274, opened 2021-06-10
Add an incomplete warning to the top page

JohnTitor:
Since we currently have a lot of outdated/missing content, it makes sense to show a warning like the reference.


Gankra (approved):
sgtm

Gankra:
needs a manual rebase, it seems

PR #276, opened 2021-06-10
Add a link to show why unused lifetimes on structs are forbidden

### LINKED ISSUE #196: Add links to troubles caused by unbounded lifetimes and types inside struct definitions
The **nomicon** page for `PhantomData` (src/phantom-data.md) contains the paragraph as follows:
> However because `'a` is unused within the struct's body, it's unbounded. **Because of the troubles this has historically caused, unbounded lifetimes and types are forbidden in struct definitions. Therefore we must somehow refer to these types in the body.** Correctly doing this is necessary to have correct variance and drop checking.

I think it would be nice to add in a few links to troubles(GitHub issues or forum discussions) that were caused by **unbounded lifetimes and types inside struct definitions**. 

If anyone is aware of relevant GitHub issues, adding them to the document would be helpful for new Rust learners.

As a Rust learner-in-progress myself, I'm not yet aware of the history of troubles that *unbounded lifetimes and types* caused in Rust, but maybe I could also try to find some relevant links to be added.

JohnTitor:
Fixes #196

### DIFF
--- a/src/phantom-data.md
+++ b/src/phantom-data.md
@@ -13,10 +13,12 @@ struct Iter<'a, T: 'a> {
 ```
 
 However because `'a` is unused within the struct's body, it's *unbounded*.
-Because of the troubles this has historically caused, unbounded lifetimes and
-types are *forbidden* in struct definitions. Therefore we must somehow refer
-to these types in the body. Correctly doing this is necessary to have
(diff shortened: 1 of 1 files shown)


ehuss (approved):
👍

PR #277, opened 2021-06-10
Clarify casting between the same size fixed ints

### LINKED ISSUE #261: i32 -> u32 conversion described as "no-op"
From here: https://doc.rust-lang.org/nomicon/casts.html

    casting between two integers of the same size (e.g. i32 -> u32) is a no-op

A signed integer can contain values an unsigned integer cannot, so even if this is true at the instruction level it seems misleading. Should maybe mention if Rust assumes 2s-complement representation?

JohnTitor:
Fixes #261

### DIFF
--- a/src/casts.md
+++ b/src/casts.md
@@ -46,6 +46,7 @@ expression, `e as U2` is not necessarily so.
 For numeric casts, there are quite a few cases to consider:
 
 * casting between two integers of the same size (e.g. i32 -> u32) is a no-op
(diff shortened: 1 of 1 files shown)


ehuss (approved):
👍  Though I think this should probably just point to the reference.

JohnTitor:
@ehuss Should we tweak this as well then? https://doc.rust-lang.org/nightly/reference/expressions/operator-expr.html#semantics

PR #278, opened 2021-06-10
Some improvements on the "subtyping" chapter

JohnTitor:
Addresses https://github.com/rust-lang/nomicon/issues/230#issuecomment-673054820. Also, marks one example `compile_fail` so that we can check the output on the book directly.


ehuss (approved):
👍

PR #284, opened 2021-06-22
Removed unecessary #[repr(C)] attribute

detly:
The #[repr(C)] attribute on the callback example is not necessary, since the type is not used from the C code.

### DIFF
--- a/src/ffi.md
+++ b/src/ffi.md
@@ -310,7 +310,6 @@ referenced Rust object.
 Rust code:
 
 ```rust,no_run
-#[repr(C)]
 struct RustObject {
     a: i32,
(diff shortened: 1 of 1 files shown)


Gankra (approved):
Hmm, yeah, there's no ABI or public API concerns here. I don't think it necessarily hurts to have it, better safe than sorry, but yeah maybe it's nice to show it's not needed.

detly:
> better safe than sorry, but yeah maybe it's nice to show it's not needed.

I can only tell you my opinion, but I found it confusing because it didn't relate to the context. But I see it's merged now, cheers!

PR #288, opened 2021-06-29
Audit `ignore` annotations

### LINKED ISSUE #283: Audit the use of `ignore` annotations
Currently, we use `ignore` annotations without comment, which makes it hard to understand why it's ignored. I saw the reference adds an explanation as a comment like `<!-- expanded code -->`, I like this idea. We should audit the use of `ignore` annotations and add comments to them on the nomicon as well.

JohnTitor:
Fixes #283

### DIFF
--- a/src/aliasing.md
+++ b/src/aliasing.md
@@ -56,6 +56,7 @@ and `output` overlap, such as `compute(&x, &mut x)`.
 
 With that input, we could get this execution:
 
+<!-- ignore: expanded code -->
 ```rust,ignore
                     //  input ==  output == 0xabad1dea
                     // *input == *output == 20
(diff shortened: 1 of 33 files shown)


ehuss (approved):
👍  I didn't examine every single example, but seems fine.

One possible option in the future is to use `{{#rustdoc_include}}` (along with mdbook's include anchors), and place the sample code in a separate directory where it can be more thoroughly tested.  I notice there are huge chunks of code that aren't tested at all, which means there is a good likelihood that it contains errors, or may become incorrect in the future.  TRPL is written this way.  A big downside is that it becomes more difficult to edit/read  the markdown source since the examples are in a separate file, so it is tough to decide if that is worth it.

JohnTitor:
> One possible option in the future is to use `{{#rustdoc_include}}` (along with mdbook's include anchors), and place the sample code in a separate directory where it can be more thoroughly tested. I notice there are huge chunks of code that aren't tested at all, which means there is a good likelihood that it contains errors, or may become incorrect in the future. TRPL is written this way. A big downside is that it becomes more difficult to edit/read the markdown source since the examples are in a separate file, so it is tough to decide if that is worth it.

Sounds good to me, at least we could use that for the examples using the `libc` crate, I believe. I'll open an issue about that not to forget and to discuss.

Issue #279, opened 2021-06-16
Move the lists of rules to the reference


JohnTitor:
As noted in https://github.com/rust-lang/nomicon/issues/230#issuecomment-858248008, we have some duplicate lists of the rules between the nomicon and the reference, and Eric and I would like to defer them to the reference.

So, we should:
- Collects the current lists
- Checks if the reference has it
  - If not, prepares a PR for it
- Removes it from the nomicon and adds a link to the reference

@ehuss is the above what you meant?

cc @rust-lang/lang-docs feel free to drop your thoughts, any suggestions are welcome!

ehuss:
Yea, that's probably reasonable, though I'm uncertain.  I don't think it's terribly bad if there is some duplication, the main concern is that we don't have the resources to keep things in sync, and in general that is a hassle for people contributing.  Unfortunately, removing it from the nomicon can make reading it a little harder since you would need to head over to the reference to find the actual information.  

I would suggest if removing stuff to do it carefully.  Make sure the information is complete in the reference, and that it doesn't gut the nomicon too much.  For example, the coercions page could remove the list of rules, but keep a high-level discussion and a good suite of examples.  A rule of thumb is that the reference generally doesn't explain the "why" of anything in the language. It doesn't explain why you would want to coerce, or when it is useful or important or necessary.  That seems like a better fit for the nomicon.

Some of the duplication I see is:

* The list of unsafe operations in https://doc.rust-lang.org/nightly/nomicon/what-unsafe-does.html
* Coercions https://doc.rust-lang.org/nightly/nomicon/coercions.html
* Casting https://doc.rust-lang.org/nightly/nomicon/casts.html

JohnTitor:
Yeah, I agree with removing may overdo cleanup. So, adding a note saying "this list may be outdated, see the reference for a more accurate list" is enough here?

Gankra:
Coercions and Casting were both very "the reference is dead, I need to be the reference" and can be happily moved to the reference. The list of unsafe/UB things I think is definitely worth mirroring in the nomicon, since it's kind of a fundamental pretext for everything we're doing, although that page is getting a bit unwieldy as we more precisely define things.

JohnTitor:
Another one about lifetime elision: https://github.com/rust-lang/nomicon/pull/306#issuecomment-908685816

PR #290, opened 2021-07-06
Move the list of coercions to the reference

JohnTitor:
Helps #279:
- Removed the list of coercions and added a link to the reference (at a glance, the reference completely covers the items in the list here, so just removed)
- Tweaked the wording on some lines
- Updated the diagnostics

### DIFF
--- a/src/coercions.md
+++ b/src/coercions.md
@@ -5,48 +5,9 @@ generally just *weakening* of types, largely focused around pointers and
 lifetimes. They mostly exist to make Rust "just work" in more cases, and are
 largely harmless.
 
-Here's all the kinds of coercion:
+For an exhaustive list of all the types of coercions, see the [Coercion types] section on the reference.
 
-Coercion is allowed between the following types:
-
-* Transitivity: `T_1` to `T_3` where `T_1` coerces to `T_2` and `T_2` coerces to
-  `T_3`
-* Pointer Weakening:
-  * `&mut T` to `&T`
-  * `*mut T` to `*const T`
-  * `&T` to `*const T`
-  * `&mut T` to `*mut T`
-* Unsizing: `T` to `U` if `T` implements `CoerceUnsized<U>`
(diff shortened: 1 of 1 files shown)


src/coercions.md
```diff
@@ -1,58 +1,16 @@
...
-largely harmless.
+Types can implicitly be coerced to change in certain contexts.
+These changes are generally just *weakening* of types, largely focused around pointers and lifetimes.
+They mostly exist to make Rust "just work" in more cases, and are largely harmless.
 
-Here's all the kinds of coercion:
+For an exhaustive list of all the types of coercions, see the [Coercion types] section on the reference.
 
-Coercion is allowed between the following types:
+## Cases where coercions do not perform
```
Gankra, line 9:
```suggestion
## Cases where coercions do not happen
```

(but I would drop the subheading altogether -- this is already a very short page now)

src/coercions.md
```diff
@@ -67,20 +25,21 @@ fn main() {
 }
 ```
 
+which fails like the following:
```
Gankra, line 28:
```suggestion
which fails as follows:
```

src/coercions.md
```diff
@@ -1,58 +1,16 @@
...
-* Array literals: `let x: [U; 10] = [e, ..]`
-* Tuple literals: `let x: (U, ..) = (e, ..)`
-* The last expression in a block: `let x: U = { ..; e }`
-
-Note that we do not perform coercions when matching traits (except for
-receivers, see below). If there is an impl for some type `U` and `T` coerces to
-`U`, that does not constitute an implementation for `T`. For example, the
-following will not type check, even though it is OK to coerce `t` to `&T` and
-there is an impl for `&T`:
+Note that we do not perform coercions when matching traits (except for receivers, see below).
```
Gankra, line 11:
```suggestion
Note that we do not perform coercions when matching traits (except for receivers, see the next page).
```

This is a per-existing error, but still should be fixed.

Gankra (approved):
just a few edits

JohnTitor:
Thanks! Applied all the review comments.

PR #296, opened 2021-07-28
Write a basic "call Rust from C" example

### LINKED ISSUE #138: Update/rewrite "Calling Rust code from C" section
This https://doc.rust-lang.org/nomicon/ffi.html#calling-rust-code-from-c
( and even in the book here: https://doc.rust-lang.org/stable/book/ch19-01-unsafe-rust.html#calling-rust-functions-from-other-languages )

has no mention of all the other info needed to accomplish this, such as: compile as a [cdylib](https://speakerdeck.com/dbrgn/calling-rust-from-c-and-java?slide=33) by adding [`crate-type = [ "cdylib" ]`](https://speakerdeck.com/dbrgn/calling-rust-from-c-and-java?slide=54) in `Cargo.toml` `[lib]` section, or using [`cbindgen`](https://github.com/eqrion/cbindgen) to [generate a header file](https://speakerdeck.com/dbrgn/calling-rust-from-c-and-java?slide=55), or the [`gcc` command line](https://speakerdeck.com/dbrgn/calling-rust-from-c-and-java?slide=56) to compile the `.cpp` file.

(**NOTE**: the below link is now invalid as the author deleted their account.)
I've attempted all the needed stuff in a working example [here](https://github.com/howaboutsynergy/reflo/tree/e8b160c223a0dfc84242e830759b0d0ac3ffb573/book/l_19_8_two).

JohnTitor:
Fix #138
I'm not sure if this is the simplest way but should make it easy to understand.


src/ffi.md
```diff
@@ -248,6 +248,72 @@ the release of these resources (especially in the case of panic).
...
+```rust
+#[no_mangle]
+pub extern "C" fn hello_from_rust() {
+    println!("Hello from Rust!");
+}
+# fn main() {}
+```
+
+The `extern "C"` makes this function adhere to the C calling convention, as discussed above in "[Foreign Calling Conventions]".
+The `no_mangle` attribute turns off Rust's name mangling, so that it is easier to link to.
```
ehuss, line 270:
```suggestion
The `no_mangle` attribute turns off Rust's name mangling, so that it has a well defined symbol to link to.
```

I realize you just moved this from below, but the way it is written seems strange to me.  With a mangled name, I don't think you'd be able to guess the symbol name at all, which makes linking mostly impossible?

src/ffi.md
```diff
@@ -248,6 +248,72 @@ the release of these resources (especially in the case of panic).
...
+
+Then, to compile Rust code as a shared library that can be called from C, add the following to your `Cargo.toml`:
+
+```toml
+[lib]
+crate-type = ["cdylib"]
+```
+
```
ehuss, line 278: Perhaps add a note here that `staticlib` is also another choice?

src/ffi.md
```diff
@@ -248,6 +248,72 @@ the release of these resources (especially in the case of panic).
...
+Run `cargo build` and you're ready to go on the Rust side.
+
+[Foreign Calling Conventions]: ffi.html#foreign-calling-conventions
```
ehuss, line 281:
```suggestion
[Foreign Calling Conventions]: ffi.md#foreign-calling-conventions
```

Using the `.md` extension makes it nicer to be able to use the links on GitHub.

ehuss:
FWIW, this chapter seems to be getting quite long.  At some point it might be worthwhile to consider breaking it up.

src/ffi.md
JohnTitor: I moved it because we have the "Callbacks from C code to Rust functions" section below. Applied your suggestion, is it okay with you now?

JohnTitor:
Sorry for the huge delay! I applied your comments.

> FWIW, this chapter seems to be getting quite long. At some point it might be worthwhile to consider breaking it up.

Sounds good, I'll do it on another PR not to make this PR large.

PR #303, opened 2021-08-27
Ordering and accesses _across_ memory locations

getreu:
Explain when orderings matter: only when accessing across multiple memory locations.

This summarizes a [discussion on Reddit about the subject](https://www.reddit.com/r/rust/comments/p9a740/question_find_the_right_ordering_for_atomics/)


JohnTitor (commented):
The explanation itself seems good but the place isn't good to me. I think the front of "Data Accesses" as a separated "Memory Orderings" section would be better (at least, inside of that section isn't good).

PR #306, opened 2021-08-30
update lifetime-elision to show what elided code under `rust_2018_idi…

guswynn:
…oms` looks like

This was a bit confusing for me to read, as it was using an old style of rust that I am not used to
Unfortunately, this lint group isn't on by default, but I think `elided_lifetimes_in_paths` may be on track to be deny-by-default in edition 2021?


JohnTitor (approved):
Nice improvement, thanks!

ehuss:
FWIW, I think it might be worthwhile to consider removing this chapter completely and just point it to https://doc.rust-lang.org/nightly/reference/lifetime-elision.html, since there is a large amount of duplication (and, for example, the reference already has this change).  If there is information here that is not in the reference, perhaps it could be moved over.

guswynn:
@ehuss oh thats a great idea, google offered this link first when I googled "rust lifetime elision" unfortunately

PR #308, opened 2021-09-05
Remove useless `unsafe`, `mut` and ptr casts in example in `send-and-sync.md`

WaffleLapkin:
Compiler even warns about unnecessary `unsafe` and unnecessary mutable variable:
- Current version [(playground)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=03cc92bd31179e9de69b4630e067f7b7)
- Proposed version [(playground)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4834051727bc676eb7b3f22f661ad556)


JohnTitor (approved):
Nice, thanks!

PR #314, opened 2021-09-27
Typo: C ABI -> C API

kelvin-273:


ehuss:
Thanks for the PR!  I think the original appears correct to me. The default for an extern block is the [C ABI](https://doc.rust-lang.org/nightly/reference/items/external-blocks.html#abi), and I think what the sentence is trying to say is that the functions are being defined with that ABI.

PureWhiteWu:
对的，ABI 的意思是 Application Binary Interface，和 API 是完全不一样的。

kelvin-273:
Ah my bad. This would be my first time learning about ABIs.
Could it help if the this was hyperlinked to the Rust Reference or would that open the flood gate to a whole lot of unnecessary cross referencing?

PureWhiteWu:
> Ah my bad. This would be my first time learning about ABIs.
> Could it help if the this was hyperlinked to the Rust Reference or would that open the flood gate to a whole lot of unnecessary cross referencing?

I think this is unnecessary. If someone does not understand these terms, they should go to Google themselves. After all, this book is not for beginners.

kelvin-273:
Fair enoungh

PR #316, opened 2021-10-04
Update rc decrement snipped

kendru:
The atomic decrement to `rc` should be performed with an ordering of `Ordering::Release` rather than `ordering::Relaxed` to interact correctly with the fence. The code in the full `Drop` implementation at the bottom of the section is correct.


JohnTitor (approved):
Good catch, thanks!

PR #318, opened 2021-10-07
doc: improve documentation of packed type layout modifier

phip1611:
~The Rust language reference (and my experience, proved by tests) says, that `packed(n)` gives you the ability to align the struct too. This bug probably comes from a past version of Rust, where this wasn't supported.~

Currently, the packed documentation doesn't mention the n-parameter.

https://doc.rust-lang.org/reference/type-layout.html

### DIFF
--- a/src/other-reprs.md
+++ b/src/other-reprs.md
@@ -119,26 +119,65 @@ assert_eq!(16, size_of::<MyReprOption<&u16>>());
 
 This optimization still applies to fieldless enums with an explicit `repr(u*)`, `repr(i*)`, or `repr(C)`.
 
-## repr(packed)
+## repr(packed(n))
 
-`repr(packed)` forces Rust to strip any padding, and only align the type to a
-byte. This may improve the memory footprint, but will likely have other negative
+`repr(packed(n))` can lower the alignment of every single struct member and thus 
+the alignment of the struct itself. The final alignment of each struct member is 
+`min(n, normal_alignment)`. Hence, `packed`/`packed(1)` aligns each field to a 
+one-byte boundary. Effectively, this strips any padding between member fields, 
+which may improve the memory footprint, but will likely have other negative 
 side-effects.
 
-In particular, most architectures *strongly* prefer values to be aligned. This
-may mean the unaligned loads are penalized (x86), or even fault (some ARM
-chips). For simple cases like directly loading or storing a packed field, the
-compiler might be able to paper over alignment issues with shifts and masks.
-However if you take a reference to a packed field, it's unlikely that the
-compiler will be able to emit code to avoid an unaligned load.
+In particular, most architectures *strongly* prefer values to be aligned. This 
+may mean the unaligned loads are penalized (x86), or even fault (some ARM chips). 
+For simple cases like directly loading or storing a packed field, the compiler 
+might be able to paper over alignment issues with shifts and masks. However, if 
+you take a reference to a packed field, it's unlikely that the compiler will be 
+able to emit code to avoid an unaligned load.
 
-[As this can cause undefined behavior][ub loads], the lint has been implemented
-and it will become a hard error.
+[As this can cause undefined behavior][ub loads], the lint has been implemented.
+With Rust 1.57 stable it is still a warning, but will become a hard error in the 
+future.
 
 `repr(packed)` is not to be used lightly. Unless you have extreme requirements,
 this should not be used.
 
-This repr is a modifier on `repr(C)` and `repr(Rust)`.
+This repr is a modifier on `repr(C)` and `repr(Rust)`. A typical use case is 
+`repr(C, packed)`, which gives you full control over the exact type layout in 
+memory.
+
+### Example
+
+The example down below shows how you can mutate data in a packed struct and safely
+handle unaligned pointers.
+
+```rust
+#[derive(Default)]
+#[repr(packed)]
+struct Foo {
+    a: u8,
+    b: u64,
+}
+
+impl Foo {
+    // safe way of creating an unaligned pointer to the field
+    fn b_mut_ptr(&mut self) -> *mut u64 {
+        core::ptr::addr_of_mut!(self.b)
+    }
+}
+
+fn main() {
+    println!("{:?}", {
+        let mut foo = Foo::default();
+        let ptr = foo.b_mut_ptr();
+        unsafe {
+            // safely write to the unaligned ptr
+            core::ptr::write_unaligned(ptr, *ptr + 1);
+            *ptr
+        }
+    });
+}
+```
 
 ## repr(align(n))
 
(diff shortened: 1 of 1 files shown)


src/other-reprs.md
```diff
@@ -112,26 +112,12 @@ assert_eq!(16, size_of::<MyReprOption<&u16>>());
 
 This optimization still applies to fieldless enums with an explicit `repr(u*)`, `repr(i*)`, or `repr(C)`.
 
-## repr(packed)
+## repr(packed(n))
 
-`repr(packed)` forces Rust to strip any padding, and only align the type to a
-byte. This may improve the memory footprint, but will likely have other negative
-side-effects.
+`repr(packed(n))` forces Rust to strip any padding, and align the type to a [`n`-byte boundary][rust-reference-type-layout],
```
ehuss, line 117: I don't believe this is quite correct.  The original of `#[repr(packed)]` is equivalent to `#[repr(packed(1))]` which does remove padding.  However, a value greater than 1 can retain some padding, so i think it isn't quite correct to say it strips *all* padding.

src/other-reprs.md
```diff
@@ -112,26 +112,12 @@ assert_eq!(16, size_of::<MyReprOption<&u16>>());
...
-However if you take a reference to a packed field, it's unlikely that the
-compiler will be able to emit code to avoid an unaligned load.
-
-[As this can cause undefined behavior][ub loads], the lint has been implemented
-and it will become a hard error.
-
-`repr(packed)` is not to be used lightly. Unless you have extreme requirements,
-this should not be used.
-
-This repr is a modifier on `repr(C)` and `repr(Rust)`.
```
ehuss, line 134: I'm not sure it is correct to remove all of this content.  Using `packed` is hazardous, and should be used with great care.  I believe the lint is still there and valid.

I believe the current state is captured in https://github.com/rust-lang/rust/issues/82523 and https://github.com/rust-lang/rust/issues/27060.  References to packed structure fields have graduated from "it's unlikely that the compiler will be able to emit code to avoid an unaligned load" to it is now a future-incompatible warning that will become an error in the future.

phip1611:
I updated the PR @ehuss . See the latest version here: https://github.com/rust-lang/nomicon/blob/20d86cd0331e1c54bfd85e25326e3e0e41fc979e/src/other-reprs.md

Any more comments?

src/other-reprs.md
```diff
@@ -112,26 +112,43 @@ assert_eq!(16, size_of::<MyReprOption<&u16>>());
 
 This optimization still applies to fieldless enums with an explicit `repr(u*)`, `repr(i*)`, or `repr(C)`.
 
-## repr(packed)
+## repr(packed(n))
 
-`repr(packed)` forces Rust to strip any padding, and only align the type to a
-byte. This may improve the memory footprint, but will likely have other negative
-side-effects.
+`repr(packed(n))` forces the Rust compiler to strip any padding between fields, and align the type, hence,
```
ehuss, line 117: This still isn't quite correct, as it may not necessarily remove padding between fields.

src/other-reprs.md
```diff
@@ -112,26 +112,43 @@ assert_eq!(16, size_of::<MyReprOption<&u16>>());
...
+the first field, at least to a [`n`-byte boundary][rust-reference-type-layout] (where `n` is a power of two).
+`repr(packed)` is equivalent to `repr(packed(1))`.
+
+This `repr`-attribute is a modifier on `repr(C)` and `repr(Rust)`, but is incompatible with `align(n)`. It may improve
+the memory footprint and is useful in scenarios, where you need exact control over the memory layout
```
ehuss, line 122:
```suggestion
the memory footprint and is useful in scenarios where you need exact control over the memory layout
```

src/other-reprs.md
```diff
@@ -112,26 +112,43 @@ assert_eq!(16, size_of::<MyReprOption<&u16>>());
...
+(such as constructing network packets).
 
 In particular, most architectures *strongly* prefer values to be aligned. This
```
ehuss, line 125: The "In particular" here is no longer clear, as it is now separated from the original "likely have other negative side-effects".  Perhaps the start of this paragraph can be rewritten to include the original warning?

Also, is there a reason to remove the comment about the compiler using shifts and masks to deal with an unaligned load?  That seems like it is useful to highlight if indeed that's how it works, since that highlights the performance penalties of using packed structures.

src/other-reprs.md
phip1611: What do you mean by that? Isn't `packed` exactly there to strip any padding between struct members? @ehuss

src/other-reprs.md
phip1611: In my opinion the old version sounded like it includes too much in-depth details. But you are right. I work on a better formulation.

src/other-reprs.md
ehuss: For example, in the following, there is still 1 byte of padding between the first and second field:

```rust
#[repr(packed(2))]
struct Foo {
    f1: u8,
    f2: u32,
}
```

phip1611: Ohh, I learned something! Thanks! I thought that only the beginning of the struct gets aligned to n and all other properties have no padding.

phip1611: @ehuss What's the difference between align(2) and packed(2) then, if packed doesn't have zero padding between the fields for N > 1?

ehuss: Sorry, I'm not sure I understand the question.  `align` and `packed` do opposite things.  `align` can increase the alignment of the type itself, and `packed` will lower it. In addition, `packed` can also change the padding between fields of a struct, as it will set the alignment of each field to the smaller of the specified alignment and the normal alignment of that field's type.  `align` does not change the alignment of each field, so won't change the field padding.

So, with the example above, `Foo` has 2-byte alignment, with 1 byte of padding between `f1` and `f2`.
If you use `align(2)`, then `Foo` has 4-byte alignment, with 3 bytes of padding between `f1` and `f2`, which is the same if you don't specify `align` at all.

src/other-reprs.md
```diff
@@ -112,26 +112,65 @@ assert_eq!(16, size_of::<MyReprOption<&u16>>());
...
+`repr(packed(n))` can lower the alignment of every single struct member and thus 
+the alignment of the struct itself. The final alignment of each struct member is 
+`min(n, normal_alignment)`. Hence, `packed`/`packed(1)` aligns each field to a 
+one-byte boundary. Effectively, this strips any padding between member fields, 
+which may improve the memory footprint, but will likely have other negative 
 side-effects.
 
-In particular, most architectures *strongly* prefer values to be aligned. This
-may mean the unaligned loads are penalized (x86), or even fault (some ARM
-chips). For simple cases like directly loading or storing a packed field, the
```
phip1611, line 123: I can remove this from the PR if you want. This was necessary so that each paragraph has roughly the same width in the markdown file.

src/other-reprs.md
```diff
@@ -143,7 +182,8 @@ never share the same cache line with each other (which may speed up certain
 kinds of concurrent code).
 
 This is a modifier on `repr(C)` and `repr(Rust)`. It is incompatible with
-`repr(packed)`.
+`repr(packed)`, but an struct with `align(n)` can wrap a struct, that is
```
phip1611, line 185: an => a

phip1611:
@ehuss  I rebased the PR branch onto the latest master branch.

PR #322, opened 2021-11-05
Replace some use of variant with covariant

chubei:


ehuss:
Thanks!  This looks good to me, but I'd like to see if someone else can weigh in.

JohnTitor (approved):
LGTM as well, thanks!

PR #325, opened 2021-12-08
Update the guidance on uninitialized data with ptr::addr_of_mut

kupiakos:
This is an impossible problem no longer! The example is copied straight from the stdlib.


JohnTitor (approved):
Good call, thanks!

PR #327, opened 2022-01-04
Fix typo / type error in FFI code example

khollbach:
There's a discrepancy between the C code and the Rust code that calls it.

The Rust callback produces a return code: `fn(fn(int) -> int, int) -> int`,
but the C callback returns nothing: `void (*)(int (*)(int), int)`.


JohnTitor (approved):
Good catch, thanks!

PR #329, opened 2022-01-08
Clarify repr(transparent) in other-reprs

nico-abram:
This clarifies that UnsafeCell is not just repr(transparent), since it does not propagate niches the same as it's inner type, and also mentions when it can be relied on by other crates. 

For some discussion on that second point, see this issue https://github.com/rust-lang/rust/issues/90435

### DIFF
--- a/src/other-reprs.md
+++ b/src/other-reprs.md
@@ -63,7 +63,7 @@ whole struct is guaranteed to be the same as that one field.
 The goal is to make it possible to transmute between the single field and the
 struct. An example of that is [`UnsafeCell`], which can be transmuted into
 the type it wraps ([`UnsafeCell`] also uses the unstable [no_niche][no-niche-pull],
-so it's ABI is not actually guaranteed to be the same when nested in other types).
+so its ABI is not actually guaranteed to be the same when nested in other types).
 
 Also, passing the struct through FFI where the inner field type is expected on
 the other side is guaranteed to work. In particular, this is necessary for `struct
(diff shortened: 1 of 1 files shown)


src/other-reprs.md
```diff
@@ -62,12 +62,17 @@ whole struct is guaranteed to be the same as that one field.
...
-the type it wraps.
+the type it wraps ([`UnsafeCell`] also uses the unstable [no_niche][no-niche-pull],
+so it's ABI is not actually guaranteed to be the same when nested in other types).
 
 Also, passing the struct through FFI where the inner field type is expected on
 the other side is guaranteed to work. In particular, this is necessary for `struct
 Foo(f32)` to always have the same ABI as `f32`.
 
+This repr is only considered part of the public ABI of a type if either the single
+field is pub, or if it's layout is documented in prose. Otherwise, the layout should
```
JohnTitor, line 73:
```suggestion
field is `pub`, or if its layout is documented in prose. Otherwise, the layout should
```

src/other-reprs.md
```diff
@@ -62,12 +62,17 @@ whole struct is guaranteed to be the same as that one field.
 
 The goal is to make it possible to transmute between the single field and the
 struct. An example of that is [`UnsafeCell`], which can be transmuted into
-the type it wraps.
+the type it wraps ([`UnsafeCell`] also uses the unstable [no_niche][no-niche-pull],
+so it's ABI is not actually guaranteed to be the same when nested in other types).
```
JohnTitor, line 66:
```suggestion
so its ABI is not actually guaranteed to be the same when nested in other types).
```

JohnTitor (approved):
LGTM with a few nits

nico-abram:
Applied suggested changes. I can squash if needed

JohnTitor:
> Applied suggested changes. I can squash if needed

Thanks! Don't worry, we can squash them when merging :)

PR #331, opened 2022-01-20
Make C code more recognizably C

XrXr:
As you may know, an empty parameter list can have [different semantics]
in C compared to a `(void)` parameter list. Also, while it's fine to
declare `main()` with an empty parameter list, it's implementation
defined as to whether it is acceptable.

Declare `main` with `(void)` so we use a standard-defined signature from
section 5.1.2.2.1 of C99.

[different semantics]: https://godbolt.org/z/fW66MsaMM


JohnTitor (approved):
Thanks!

PR #334, opened 2022-01-30
Make the Vec impl be slightly more careful with ZSTs and alignment.

Gankra:
This also incidentally makes the ZST code and final code have the same formatting for the divide.

### DIFF
--- a/src/vec/vec-final.md
+++ b/src/vec/vec-final.md
@@ -235,21 +235,22 @@ impl<T> Iterator for RawValIter<T> {
             None
         } else {
             unsafe {
-                let result = ptr::read(self.start);
-                self.start = if mem::size_of::<T>() == 0 {
-                    (self.start as usize + 1) as *const _
+                if mem::size_of::<T>() == 0 {
+                    self.start = (self.start as usize + 1) as *const _;
+                    Some(ptr::read(NonNull::<T>::dangling().as_ptr()))
                 } else {
-                    self.start.offset(1)
-                };
-                Some(result)
+                    let old_ptr = self.start;
+                    self.start = self.start.offset(1);
+                    Some(ptr::read(old_ptr))
+                }
             }
         }
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
         let elem_size = mem::size_of::<T>();
-        let len = (self.end as usize - self.start as usize) /
-                  if elem_size == 0 { 1 } else { elem_size };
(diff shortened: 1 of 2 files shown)


src/vec/vec-final.md
```diff
@@ -235,21 +235,22 @@ impl<T> Iterator for RawValIter<T> {
...
+                    self.start = (self.start as usize + 1) as *const _;
+                    Some(ptr::read(NonNull::<T>::dangling().as_ptr()))
                 } else {
-                    self.start.offset(1)
-                };
-                Some(result)
+                    let old_ptr = self.start;
+                    self.start = self.start.offset(1);
+                    Some(ptr::read(old_ptr))
+                }
```
eddyb, line 245: Would it make sense to factor out a single raw pointer that is `Some(ptr::read(elem_ptr))`'d *afterwards*?

Gankra: I think it's very "doesn't matter" because this is a static branch, but I am mildly more inclined to this impl because it's trivial to swap it out with the read_unaligned version if you want that instead

Gankra: Or I guess put another way: these are two wholy distinct implementations that we're shoveling into one function body, and trying to reuse code was the reason we had a bug in the first place, so let's just be more honest about that reality and not mix the code.

src/vec/vec-zsts.md
```diff
@@ -130,7 +130,41 @@ Now we have a different bug. Instead of our iterators not running at all, our
...
+Do you see the "bug"? No one else did! The original author only noticed the
+problem when linking to this page years later. This code is kind of dubious
+because abusing the iterator pointers to be *counters* makes them unaligned!
+Our *one job* when using ZSTs is to keep pointers aligned! *forehead slap*
+
+Raw pointers don't need to be aligned at all times, so the basic trick of
+using pointers as counters is *fine*, but they *should* definitely be aligned
+when passed to `ptr::read`! This is *possibly* needless pedantry
+because `ptr::read` is a noop for a ZST, but let's be a *little* more
+responsible and read from NonNull::danging on the ZST path.
```
miscco, line 163: Typo here, although `danging` is remarkably fitting

Gankra:
Should be ready to land, just needs approval

JohnTitor (approved):
LGTM, @eddyb could you re-review this if you have some time?

PR #335, opened 2022-02-02
Fix a syntax error in leaking.md

BillXWB:
`guards` should be mutable here:
https://github.com/rust-lang/nomicon/blob/9493715a6280a1f74be759c7e1ef9999b5d13e6f/src/leaking.md#L208-L232


ehuss (approved):
Thanks!

Issue #339, opened 2022-02-09
The Subtyping and Variance chapter is confusing


mejrs:
Today in the Rust community discord we had a discussion about variance and the nomicon chapter discussing it (starting [here](https://discord.com/channels/273534239310479360/818964227783262209/941083578421354526)), and how it's generally confusing and is something that just doesn't "click" for a lot of people (myself included).

# First impression 

I'll start off by summarizing my (and others') impressions as we read the chapter. 

> Subtyping is a relationship between types that allows statically typed languages to be a bit more flexible and permissive.

Uh, ok. This tells nothing about what subtyping is supposed to mean. (it's much later that this is actually explained)

> Subtyping in Rust is a bit different from subtyping in other languages.

Well I don't know what "subtyping in other languages" is, either.

> To keep things simple, this section will consider a small extension to the Rust language....
> ...So here's our simple extension, Objective Rust

Seems kinda random how we got here. Why are we introducing a new language? I'm interested in what variance is in *Rust* and why I should care about it. This is anything but simple.

> <lots of cats, dogs and animals>

At this point the article has completely lost me.

# Constructive feedback

There are some things that could be done better:

 - Show locked doors before we show keys. We probably want to introduce basic subtyping (you can pass `&'static Foo` to `fn bar<'a>(&'a Foo))`, then show an issue with doing that (adapt the "meowing dogs" example to lifetimes) and then introduce variance to deal with this.
 - explain "is a subtype of" in terms of "is as useful as and more"
 - Possibly make readers read less cats and dogs before explaining it again in terms of lifetimes.
 - I think the author(s) underestimate how unintuitive cats, dogs and animals analogies actually are (see also the linked article below). Please don't use inheritance-y analogies.
- The article dumps a lot of "we will explain this later" on the reader. Please don't underestimate how exhausting and confusing this is. For example there's an enormous distance between when variance is first mentioned and when it is explained.

Finally I would recommend you to read https://viralinstruction.com/posts/defense/. The author's frustration with teaching classes and inheritance is how I feel w/r to subtyping and variance.

PR #340, opened 2022-02-10
Rewrite the chapter on subtyping and variance

conradludgate:
This is to address the issues raised in #339

### DIFF
--- a/src/subtyping.md
+++ b/src/subtyping.md
@@ -7,34 +7,7 @@ or permit undefined behavior.
 In order to allow flexible usage of lifetimes
 while also preventing their misuse, Rust uses a combination of **Subtyping** and **Variance**.
 
-## Subtyping
-
-Subtyping is the idea that one type can be used in place of another.
-
-Let's define that `Sub` is a subtype of `Super` (we'll be using the notation `Sub: Super` throughout this chapter)
-
-What this is suggesting to us is that the set of *requirements* that `Super` defines
-are completely satisfied by `Sub`. `Sub` may then have more requirements.
-
-An example of simple subtyping that exists in the language are [supertraits](https://doc.rust-lang.org/stable/book/ch19-03-advanced-traits.html?highlight=supertraits#using-supertraits-to-require-one-traits-functionality-within-another-trait)
-
-```rust
-use std::fmt;
-
-pub trait Error: fmt::Display {
-    fn source(&self) -> Option<&(dyn Error + 'static)>;
-    fn description(&self) -> &str;
-    fn cause(&self) -> Option<&dyn Error>;
-}
-```
-
-Here, we have that `Error: fmt::Display` (`Error` is a *subtype* of `Display`),
-because it has all the requirements of `fmt::Display`, plus the `source`/`description`/`cause` functions.
-
-However, subtyping in traits is not that interesting in the case of Rust.
-Here in the nomicon, we're going to focus more with how subtyping interacts with **lifetimes**
-
-Take this example
+Let's start with a example.
 
 ```rust
 fn debug<T: std::fmt::Debug>(a: T, b: T) {
@@ -68,16 +41,53 @@ This would be rather unfortunate. In this case,
 what we want is to accept any type that lives *at least as long* as `'b`.
 Let's try using subtyping with our lifetimes.
 
-Let's define a lifetime to have the a simple set of requirements:
+## Subtyping
+
+Subtyping is the idea that one type can be used in place of another.
+
+Let's define that `Sub` is a subtype of `Super` (we'll be using the notation `Sub: Super` throughout this chapter)
+
+What this is suggesting to us is that the set of *requirements* that `Super` defines
+are completely satisfied by `Sub`. `Sub` may then have more requirements.
+
+An example of simple subtyping that exists in the language are [supertraits](https://doc.rust-lang.org/stable/book/ch19-03-advanced-traits.html?highlight=supertraits#using-supertraits-to-require-one-traits-functionality-within-another-trait)
+
+```rust
+use std::fmt;
+
+pub trait Error: fmt::Display {
+    fn source(&self) -> Option<&(dyn Error + 'static)>;
+    fn description(&self) -> &str;
+    fn cause(&self) -> Option<&dyn Error>;
+}
+```
+
+Here, we have that `Error: fmt::Display` (`Error` is a *subtype* of `Display`),
+because it has all the requirements of `fmt::Display`, plus the `source`/`description`/`cause` functions.
+
+However, subtyping in traits is not that interesting.
+Here in the nomicon, we're going to focus more with how subtyping interacts with lifetimes
+
+Let's define a lifetime to be the simple requirement:
 `'a` defines a region of code in which a value will be alive.
 Now that we have a defined set of requirements for lifetimes, we can define how they relate to each other.
-`'a: 'b` if and only if `'a` defines a region of code that **completely contains** `'b`.
+`'long: 'short` if and only if `'long` defines a region of code that **completely contains** `'short`.
 
-`'a` may define a region larger than `'b`, but that still fits our definition.
+`'long` may define a region larger than `'short`, but that still fits our definition.
+
+> As we will see throughout the rest of this chapter,
+subtyping is a lot more complicated and subtle than this,
+but this simple rule is a very good 99% intuition.
+And unless you write unsafe code, the compiler will automatically handle all the corner cases for you.
+
+> But this is the Rustonomicon. We're writing unsafe code,
+so we need to understand how this stuff really works, and how we can mess it up.
+
 Going back to our example above, we can say that `'static: 'b`.
-
-For now, let's accept the idea that subtypes of lifetimes can be passed through references (more on this in [Variance](#variance)),
-eg. `&'static str` is a subtype of `&'b str`, then we can let them coerce, and then the example above will compile
+For now, let's also accept the idea that subtypes of lifetimes can be passed through references
+(more on this in [Variance](#variance)),
+eg. `&'static str` is a subtype of `&'b str`, then we can let them coerce,
+and then the example above will compile
 
 ```rust
 fn debug<T: std::fmt::Debug>(a: T, b: T) {
--- a/src/subtyping.md
+++ b/src/subtyping.md
@@ -12,7 +12,7 @@ Let's start with an example.
 ```rust
 // Note: debug expects two parameters with the *same* lifetime
 fn debug<'a>(a: &'a str, b: &'a str) {
-    println!("a = {:?} b = {:?}", a, b);
+    println!("a = {a:?} b = {b:?}");
 }
 
 fn main() {
@@ -25,7 +25,7 @@ fn main() {
 }
 ```
 
-In a conservative implementation of lifetimes, since `hello` and `world` have differing lifetimes,
+In a conservative implementation of lifetimes, since `hello` and `world` have different lifetimes,
 we might see the following error:
 
 ```text
@@ -72,12 +72,12 @@ so we need to understand how this stuff really works, and how we can mess it up.
 Going back to our example above, we can say that `'static <: 'world`.
 For now, let's also accept the idea that subtypes of lifetimes can be passed through references
 (more on this in [Variance](#variance)),
-_e.g._ `&'static str` is a subtype of `&'world str`, then we can let a `&'static str` "downgrade" into a `&'world str`.
+_e.g._ `&'static str` is a subtype of `&'world str`, then we can "downgrade" `&'static str` into a `&'world str`.
 With that, the example above will compile:
 
 ```rust
-fn debug<T: std::fmt::Debug>(a: T, b: T) {
-    println!("a = {:?} b = {:?}", a, b);
+fn debug<'a>(a: &'a str, b: &'a str) {
+    println!("a = {a:?} b = {b:?}");
 }
 
 fn main() {
@@ -95,7 +95,7 @@ fn main() {
 Above, we glossed over the fact that `'static <: 'b` implied that `&'static T <: &'b T`. This uses a property known as _variance_.
 It's not always as simple as this example, though. To understand that, let's try to extend this example a bit:
 
-```rust,compile_fail
+```rust,compile_fail,E0597
 fn assign<T>(input: &mut T, val: T) {
     *input = val;
 }
@@ -106,7 +106,7 @@ fn main() {
         let world = String::from("world");
         assign(&mut hello, &world);
     }
-    println!("{}", hello); // use after free 😿
+    println!("{hello}"); // use after free 😿
 }
 ```
 
@@ -177,7 +177,7 @@ For more types, see the ["Variance" section][variance-table] on the reference.
 Now that we have some more formal understanding of variance,
 let's go through some more examples in more detail.
 
-```rust,compile_fail
+```rust,compile_fail,E0597
 fn assign<T>(input: &mut T, val: T) {
     *input = val;
 }
@@ -188,7 +188,7 @@ fn main() {
         let world = String::from("world");
         assign(&mut hello, &world);
     }
-    println!("{}", hello);
+    println!("{hello}");
 }
 ```
 
@@ -230,7 +230,7 @@ This is counter to the `&T` case:
 
 ```rust
 fn debug<T: std::fmt::Debug>(a: T, b: T) {
-    println!("a = {:?} b = {:?}", a, b);
+    println!("a = {a:?} b = {b:?}");
 }
 ```
 
--- a/src/subtyping.md
+++ b/src/subtyping.md
@@ -244,7 +244,7 @@ All it does is take a mutable reference and a value and overwrite the referent w
 What's important about this function is that it creates a type equality constraint. It
 clearly says in its signature the referent and the value must be the *exact same* type.
 
-Meanwhile, in the caller we pass in `&mut &'static str` and `&'spike_str str`.
+Meanwhile, in the caller we pass in `&mut &'static str` and `&'world str`.
 
 Because `&mut T` is invariant over `T`, the compiler concludes it can't apply any subtyping
 to the first argument, and so `T` must be exactly `&'static str`.
@@ -263,9 +263,9 @@ So the compiler decides that `&'static str` can become `&'b str` if and only if
 `&'static str` is a subtype of `&'b str`, which will hold if `'static: 'b`.
 This is true, so the compiler is happy to continue compiling this code.
 
-As it turns out, the argument for why it's ok for Box (and Vec, Hashmap, etc.) to be covariant is pretty similar to the argument for why it's ok for lifetimes to be covariant: as soon as you try to stuff them in something like a mutable reference, they inherit invariance and you're prevented from doing anything bad.
+As it turns out, the argument for why it's ok for Box (and Vec, HashMap, etc.) to be covariant is pretty similar to the argument for why it's ok for lifetimes to be covariant: as soon as you try to stuff them in something like a mutable reference, they inherit invariance and you're prevented from doing anything bad.
 
-However Box makes it easier to focus on by-value aspect of references that we partially glossed over.
+However Box makes it easier to focus on the by-value aspect of references that we partially glossed over.
 
 Unlike a lot of languages which allow values to be freely aliased at all times, Rust has a very strict rule: if you're allowed to mutate or move a value, you are guaranteed to be the only one with access to it.
 
@@ -279,7 +279,7 @@ world = hello;
 ```
 
 There is no problem at all with the fact that we have forgotten that `hello` was alive for `'static`,
-because as soon as we moved `hello` to a variable that only knew he it was alive for `'b`,
+because as soon as we moved `hello` to a variable that only knew it was alive for `'b`,
 **we destroyed the only thing in the universe that remembered it lived for longer**!
 
 Only one thing left to explain: function pointers.
--- a/src/subtyping.md
+++ b/src/subtyping.md
@@ -2,11 +2,41 @@
 
 Rust uses lifetimes to track the relationships between borrows and ownership.
 However, a naive implementation of lifetimes would be either too restrictive,
-or permit undefined behaviour. Let's see a few examples:
+or permit undefined behavior.
 
-```rust,ignore
-fn debug<'a>(a: &'a str, b: &'a str) {
-    println!("a = {:?} b = {:?}", a, b)
+In order to allow flexible usage of lifetimes
+while also preventing mis-use, Rust uses a combination of **Subtyping** and **Variance**.
+
+## Subtyping
+
+Subtyping is the idea that one type can be a *subtype* of another.
+Let's define that `A: B` is equivalent to saying '`A` is a subtype of `B`'.
+What this is suggesting to us is that the set of *requirements* that `B` defines
+are completely satisfied by `A`. `A` may then have more requirements.
+
+An example of simple subtyping that exists in the language are [supertraits](https://doc.rust-lang.org/stable/book/ch19-03-advanced-traits.html?highlight=supertraits#using-supertraits-to-require-one-traits-functionality-within-another-trait)
+
+```rust
+use std::fmt;
+
+trait OutlinePrint: fmt::Display {
+    fn outline_print(&self) {
+        todo!()
+    }
+}
+```
+
+Here, we have that `OutlinePrint: fmt::Display` (`OutlinePrint` is a *subtype* of `Display`),
+because it has all the requirements of `fmt::Display`, plus the `outline_print` function.
+
+However, subtyping in traits is not that interesting in the case of Rust.
+Here in the nomicon, we're going to focus more with how subtyping interacts with **lifetimes**
+
+Take this example
+
+```rust
+fn debug<T: std::fmt::Debug>(a: T, b: T) {
+    println!("a = {:?} b = {:?}", a, b);
 }
 
 fn main() {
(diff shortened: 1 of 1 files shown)


src/subtyping.md
```diff
@@ -1,146 +1,127 @@
...
+while also preventing their misuse, Rust uses a combination of **Subtyping** and **Variance**.
 
-To keep things simple, this section will consider a small extension to the
-Rust language that adds a new and simpler subtyping relationship. After
-establishing concepts and issues under this simpler system,
-we will then relate it back to how subtyping actually occurs in Rust.
+## Subtyping
 
-So here's our simple extension, *Objective Rust*, featuring three new types:
+Subtyping is the idea that one type can be a *subtype* of another.
```
mejrs, line 12:
```suggestion
Subtyping is the idea that one type can be used in place of another.
```

I'm not happy with the current line, because a reader probably doesn't know what a subtype is either.

src/subtyping.md
```diff
@@ -1,146 +1,127 @@
...
+
+```rust
+use std::fmt;
+
+trait OutlinePrint: fmt::Display {
+    fn outline_print(&self) {
+        todo!()
+    }
 }
 ```
```
mejrs, line 29: I think the [`Error`](https://doc.rust-lang.org/std/error/trait.Error.html) trait probably fits better here.

- It already exists in the standard library.
- It's a trait users will have used (or at least seen), and maybe even implemented.

When I see people asking about "traits requiring other traits" it is usually the example I reach for, because it's very obvious - In order to use something as an error, it must be displayable..

src/subtyping.md
```diff
@@ -1,146 +1,127 @@
...
+Here in the nomicon, we're going to focus more with how subtyping interacts with **lifetimes**
+
+Take this example
 
-<!-- ignore: simplified code -->
 ```rust,ignore
-fn love(pet: Animal) {
-    pet.snuggle();
+fn debug<T: std::fmt::Debug>(a: T, b: T) {
+    println!("a = {:?} b = {:?}", a, b);
```
Nilstrieb, line 41:
```suggestion
    println!("a = {a:?} b = {b:?}");
```
This raises the MSRV of the nomicon examples, but I think we can take that :). Same for the other examples.

src/subtyping.md
```diff
@@ -1,146 +1,127 @@
...
 
-But unlike normal traits, we can use them as concrete and sized types, just like structs.
+Here, we have that `OutlinePrint: fmt::Display` (`OutlinePrint` is a *subtype* of `Display`),
+because it has all the requirements of `fmt::Display`, plus the `outline_print` function.
 
-Now, say we have a very simple function that takes an Animal, like this:
+However, subtyping in traits is not that interesting in the case of Rust.
+Here in the nomicon, we're going to focus more with how subtyping interacts with **lifetimes**
+
+Take this example
```
mejrs, line 37: Personally I would lead the chapter with an example, *before even mentioning variance and subtyping*, showing an example that should compile but doesn't, where the user can see that it makes sense, and then drop *we can fix this with variance*

Nilstrieb (commented):
I really liked the rewrite, it's shorter and clearer, explaining everything without cats and dogs.

Small suggestion: I'd use the cool new formatting syntax

src/subtyping.md
```diff
@@ -1,146 +1,127 @@
...
-As we will see throughout the rest of this section, subtyping is a lot more complicated
-and subtle than this, but this simple rule is a very good 99% intuition. And unless you
-write unsafe code, the compiler will automatically handle all the corner cases for you.
+```rust
+fn debug<T: std::fmt::Debug>(a: T, b: T) {
+    println!("a = {:?} b = {:?}", a, b);
+}
 
-But this is the Rustonomicon. We're writing unsafe code, so we need to understand how
-this stuff really works, and how we can mess it up.
```
mejrs, line 74: I actually really liked this section, can we keep it? I think it's important to say this stuff is important when writing unsafe. And maybe include an example of messing it up?

mejrs (commented):
Thanks, this is much better :)

src/subtyping.md
conradludgate: Absolutely. It must have slipped past me while I was cleaning it up

src/subtyping.md
```diff
@@ -152,38 +133,36 @@ A type constructor F's *variance* is how the subtyping of its inputs affects the
...
 |   | `*const T`      |           | covariant         |           |
 |   | `*mut T`        |           | invariant         |           |
 
-The types with \*'s are the ones we will be focusing on, as they are in
-some sense "fundamental". All the others can be understood by analogy to the others:
+Some of these can be explained simply in relation to the others:
 
 * `Vec<T>` and all other owning pointers and collections follow the same logic as `Box<T>`
 * `Cell<T>` and all other interior mutability types follow the same logic as `UnsafeCell<T>`
+* `UnsafeCell<T>` having interior mutability gives it the same variance properties as `&mut T`
```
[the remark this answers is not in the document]
conradludgate: Can you think of a good way to explain that?

The original mostly just says that since `UnsafeCell<T>` allows you to treat a `&T` as a `&mut T`, it should have the same invariance

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
-and especially variance, is already hard to understand properly. As in,
-even compiler writers mess it up all the time.
+In order to allow flexible usage of lifetimes
+while also preventing their misuse, Rust uses a combination of **Subtyping** and **Variance**.
 
-To keep things simple, this section will consider a small extension to the
-Rust language that adds a new and simpler subtyping relationship. After
-establishing concepts and issues under this simpler system,
-we will then relate it back to how subtyping actually occurs in Rust.
+Let's start with a example.
```
nrc, line 10:
```suggestion
Let's start with an example.
```

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
+        let world = &world; // 'b has a shorter lifetime than 'static
+        debug(hello, world);
+    }
 }
+```
 
-trait Dog: Animal {
-    fn bark(&self);
-}
+In a conservative implementation of lifetimes, since `a` and `b` have differeing lifetimes,
```
nrc, line 27:
```suggestion
In a conservative implementation of lifetimes, since `a` and `b` have differing lifetimes,
```

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
-By default, static types must match *exactly* for a program to compile. As such,
-this code won't compile:
+Let's define that `Sub` is a subtype of `Super` (we'll be using the notation `Sub: Super` throughout this chapter)
 
-<!-- ignore: simplified code -->
-```rust,ignore
-let mr_snuggles: Cat = ...;
-love(mr_snuggles);         // ERROR: expected Animal, found Cat
-```
+What this is suggesting to us is that the set of *requirements* that `Super` defines
```
nrc, line 50: I'm not sure if requirements is the best word here. Perhaps it is better to use terminology like "A value which has type Sub can be used in any way that a value of type Super can, and possibly more"

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
+It shouldn't be surprising that we might want to assign a `T` into a `T`.
 
-Variance is where things get a bit complicated.
+The problem is that we cannot assume that `&mut &'static str` and `&mut &'b str` are compatible.
+This must mean that `&mut &'static str` should **not** be a *subtype* of `&mut &'b str`,
+even if `'static` is a subtype of `'b`.
 
-Variance is a property that *type constructors* have with respect to their
-arguments. A type constructor in Rust is any generic type with unbound arguments.
+Variance is the way that Rust defines the relationships of subtypes through their *type constructor*.
```
nrc, line 139: Variance is not just a Rust concept

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
-```rust,ignore
-fn love(pet: Animal) {
-    pet.snuggle();
-}
-```
+Subtyping is the idea that one type can be used in place of another.
 
-By default, static types must match *exactly* for a program to compile. As such,
-this code won't compile:
+Let's define that `Sub` is a subtype of `Super` (we'll be using the notation `Sub: Super` throughout this chapter)
```
LegionMammal978, line 48:
```suggestion
Let's define that `Sub` is a subtype of `Super` (we'll be using the notation `Sub: Super` throughout this chapter).
```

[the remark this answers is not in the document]
conradludgate: Aren't these two the same thing? `'a` outlives `'b` means that `'a` is a subtype of `'b`

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
-<!-- ignore: simplified code -->
-```rust,ignore
-let mr_snuggles: Cat = ...;
-love(mr_snuggles);         // ERROR: expected Animal, found Cat
-```
+What this is suggesting to us is that the set of *requirements* that `Super` defines
+are completely satisfied by `Sub`. `Sub` may then have more requirements.
 
-Mr. Snuggles is a Cat, and Cats aren't *exactly* Animals, so we can't love him! 😿
+An example of simple subtyping that exists in the language are [supertraits](https://doc.rust-lang.org/stable/book/ch19-03-advanced-traits.html?highlight=supertraits#using-supertraits-to-require-one-traits-functionality-within-another-trait)
```
LegionMammal978, line 53:
```suggestion
An example of simple subtyping that exists in the language is [supertraits](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-supertraits-to-require-one-traits-functionality-within-another-trait):
```
It might also be useful to put the URL at the end using an anchor, as in `send-and-sync.md`.

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
-```rust,ignore
-fn evil_feeder(pet: &mut Animal) {
-    let spike: Dog = ...;
+> But this is the Rustonomicon. We're writing unsafe code,
+so we need to understand how this stuff really works, and how we can mess it up.
+
+Going back to our example above, we can say that `'static: 'b`.
+For now, let's also accept the idea that subtypes of lifetimes can be passed through references
+(more on this in [Variance](#variance)),
+eg. `&'static str` is a subtype of `&'b str`, then we can let them coerce,
```
LegionMammal978, line 89:
```suggestion
e.g. `&'static str` is a subtype of `&'b str`, then we can let them coerce,
```

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
 
-Lifetimes are just regions of code, and regions can be partially ordered with the *contains*
-(outlives) relationship. Subtyping on lifetimes is in terms of that relationship:
-if `'big: 'small` ("big contains small" or "big outlives small"), then `'big` is a subtype
-of `'small`. This is a large source of confusion, because it seems backwards
-to many: the bigger region is a *subtype* of the smaller region. But it makes
-sense if you consider our Animal example: Cat is an Animal *and more*,
-just as `'big` is `'small` *and more*.
+Above, we glossed over the fact that `'static: 'b` implied that `&'static T: &'b T`. This uses a property known as variance.
+It's not always as simple as this example though, to understand that let's try extend this example a bit
```
LegionMammal978, line 110:
```suggestion
It's not always as simple as this example, though. To understand that, let's try to extend this example a bit:
```

src/subtyping.md
```diff
@@ -320,99 +244,117 @@ Meanwhile, in the caller we pass in `&mut &'static str` and `&'spike_str str`.
 Because `&mut T` is invariant over `T`, the compiler concludes it can't apply any subtyping
 to the first argument, and so `T` must be exactly `&'static str`.
 
-The other argument is only an `&'a str`, which *is* covariant over `'a`. So the compiler
-adopts a constraint: `&'spike_str str` must be a subtype of `&'static str` (inclusive),
-which in turn implies `'spike_str` must be a subtype of `'static` (inclusive). Which is to say,
-`'spike_str` must contain `'static`. But only one thing contains `'static` -- `'static` itself!
+This is counter to the `&T` case
```
LegionMammal978, line 247:
```suggestion
This is counter to the `&T` case:
```

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
+10 |         debug(hello, world);
+   |                      ^
+   |                      |
+   |                      expected `&'static str`, found struct `&'b str`
 ```
 
-But unlike normal traits, we can use them as concrete and sized types, just like structs.
+This would be rather unfortunate. In this case,
+what we want is to accept any type that lives *at least as long* as `'b`.
+Let's try using subtyping with our lifetimes.
```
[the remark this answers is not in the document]
conradludgate: Hmm. This is contradictory to how the original page represents `&'a T` as a type constructor with type arguments `'a` and `T`

src/subtyping.md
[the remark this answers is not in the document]
nrc: I don't think outlives is a subtype relation because lifetimes are not types, it could be called sub-lifetimes, but we don't call it that.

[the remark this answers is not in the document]
jackh726: > Aren't these two the same thing? `'a` outlives `'b` means that `'a` is a subtype of `'b`

That's actually the opposite of what's true: `'a: 'b` means `'a <: 'b`

jyn514 (approved):
I like the idea of using lifetimes to start instead of the cat + dog examples :) I only skimmed the PR, but it would be nice to have an example of why `fn borrow<'a>(x: &'a Chunk<'a>) {}` works (usually) but `fn borrow_mut<'a>(x: &'a mut Chunk<'a>) {}` is almost impossible to use. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1eed7dc7b916058c4a06a7da908717e4

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
-As we will see throughout the rest of this section, subtyping is a lot more complicated
-and subtle than this, but this simple rule is a very good 99% intuition. And unless you
-write unsafe code, the compiler will automatically handle all the corner cases for you.
+However, subtyping in traits is not that interesting.
+Here in the nomicon, we're going to focus more with how subtyping interacts with lifetimes
 
-But this is the Rustonomicon. We're writing unsafe code, so we need to understand how
-this stuff really works, and how we can mess it up.
+Let's define a lifetime to be the simple requirement:
+`'a` defines a region of code in which a value will be alive.
```
danielhenrymantilla, line 72:
```suggestion
`'a` defines a region of code.
```

When part of a borrow, it thus represents a region of code within which a value is borrowed. A fortiori, the region within which that borrowee is alive needs to be at least as big as the region of the borrow.

In other words, _lifetime_ is a misnomer; while it could have made sense back in the day with lexical lifetimes (and even then it would have been a stretch), now with NLL it doesn't really match either the life-time of the borrowee, nor that of the borrower:

```rs
/* 0 */ {
/* 1 */     let mut x = 42;
/* 2 */     {
/* 3 */         let r = &x;
/* 4 */         println!("{}", *r);
/* 5 */         x = 0;
/* 6 */         println!("{}", x);
/* 7 */     }
/* 8 */     println!("{}", x);
/* 9 */ }
```

  - The _alive_-time of `x` is `1.. 9`
  - The _alive_-time of `r` is `3.. 7`
  - The _lifetime_ / span of the borrow of the `&x` borrow (held by `r`) is `3.. 5`

Before NLL, the second and third spans had to match; they don't since.

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
+Now that we have a defined set of requirements for lifetimes, we can define how they relate to each other.
+`'long: 'short` if and only if `'long` defines a region of code that **completely contains** `'short`.
```
danielhenrymantilla, line 74:
```suggestion
`'long : 'short` if and only if `'long` defines a region of code that **completely contains** `'short` (_i.e._, `'long ⊇ 'short`)
```

I've added the whitespace to `'a : 'b` kind of bounds, since we're not really within rustfmt-ed code anymore, we're teaching stuff, and `:` for lifetimes represents an (antisymmetrical) binary "boolean" operator, so there is no reason to favor the left hand side over the right hand side.

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
+And unless you write unsafe code, the compiler will automatically handle all the corner cases for you.
 
-<!-- ignore: simplified code -->
-```rust,ignore
-fn evil_feeder(pet: &mut Animal) {
-    let spike: Dog = ...;
+> But this is the Rustonomicon. We're writing unsafe code,
+so we need to understand how this stuff really works, and how we can mess it up.
+
+Going back to our example above, we can say that `'static: 'b`.
```
danielhenrymantilla, line 86:
```suggestion
Going back to our example above, we can say that `'static : 'b`, as mentioned just above.
```

or, if my previous suggestions were not to be added:

```suggestion
Going back to our example above, we can say that `'static: 'b`. Indeed `'static` stands for `'forever`: the never-ending / infinite region. Since any other region is then thus contained within it (`'static ⊇ 'b`), we have `'static : 'b` for any lifetime `'b`.
```

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
-this stuff really works, and how we can mess it up.
+Let's define a lifetime to be the simple requirement:
+`'a` defines a region of code in which a value will be alive.
+Now that we have a defined set of requirements for lifetimes, we can define how they relate to each other.
+`'long: 'short` if and only if `'long` defines a region of code that **completely contains** `'short`.
 
-The core problem is that this rule, naively applied, will lead to *meowing Dogs*. That is,
-we can convince someone that a Dog is actually a Cat. This completely destroys the fabric
-of our static type system, making it worse than useless (and leading to Undefined Behaviour).
+`'long` may define a region larger than `'short`, but that still fits our definition.
```
danielhenrymantilla, line 76:
```suggestion
`'long` may define a region larger than `'short`, but that still fits our definition.

### Some basic properties

For any lifetimes `'a`, `'b`, and `'c`:

  - `'a : 'a` (a region fits inside itself / `'a = 'a ⇒ 'a ⊇ 'a`);

  - if `'a : 'b` and `'b : 'a`, then `'a = 'b` (the only way for two regions to be inside one another is if they're equal / `('a ⊇ 'b & 'b ⊇ 'a) ⇒ 'a = 'b`);

  - if `'a : 'b`, and `'b : 'c`, then `'a : 'c` (if `'b` is part of `'a`, and `'c` is part of `'b`, then necessarily `'c` is part of `'a` / `('a ⊇ 'b & 'b ⊇ 'c) ⇒ 'c ⊇ 'a`.

  - `'static` stands for `'forever`, the never-ending / infinite region. Thus `'static ⊇ 'a` for any `'a`; that is:
      `'static : 'a`.
```

This serves both as concrete pedagogical examples, and genuine tools to be used later on 🙂

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
 
-Here's a simple example of this happening when we apply subtyping in a completely naive
-"find and replace" way.
+> As we will see throughout the rest of this chapter,
+subtyping is a lot more complicated and subtle than this,
+but this simple rule is a very good 99% intuition.
+And unless you write unsafe code, the compiler will automatically handle all the corner cases for you.
```
danielhenrymantilla, line 81:
```suggestion
And unless you write unsafe code, the compiler will automatically handle all the corner cases for you.

  - Note that the aforementioned properties (`:` between lifetimes defines a partial order with `'static` being the maximum) remain true even the "region" representation falls apart.
```

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
-<!-- ignore: simplified code -->
-```rust,ignore
-fn evil_feeder(pet: &mut Animal) {
-    let spike: Dog = ...;
+> But this is the Rustonomicon. We're writing unsafe code,
+so we need to understand how this stuff really works, and how we can mess it up.
+
+Going back to our example above, we can say that `'static: 'b`.
+For now, let's also accept the idea that subtypes of lifetimes can be passed through references
+(more on this in [Variance](#variance)),
```
danielhenrymantilla, line 88:
```suggestion
For now, let's also accept the idea that subtypes of lifetimes can be passed through references
_i.e._, "the lifetime of borrow is allowed to shrink" (more on this in [Variance](#variance)),
```

src/subtyping.md
danielhenrymantilla, line 89: _coerce_ is not really a great word, here, since it does have a specific meaning in Rust which subtyping is distinct from.

 -  Maybe something along the lines of?

    ```suggestion
    _e.g._, `&'static str` is a subtype of `&'b str`, then we can let a `&'static str` "become" a `&'b str`.
    ```

    🤷

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
+## Variance
 
-Lifetimes are just regions of code, and regions can be partially ordered with the *contains*
-(outlives) relationship. Subtyping on lifetimes is in terms of that relationship:
-if `'big: 'small` ("big contains small" or "big outlives small"), then `'big` is a subtype
-of `'small`. This is a large source of confusion, because it seems backwards
-to many: the bigger region is a *subtype* of the smaller region. But it makes
-sense if you consider our Animal example: Cat is an Animal *and more*,
-just as `'big` is `'small` *and more*.
+Above, we glossed over the fact that `'static: 'b` implied that `&'static T: &'b T`. This uses a property known as variance.
```
danielhenrymantilla, line 109:
```suggestion
Above, we glossed over the fact that `'static : 'b` implied that `&'static T <: &'b T`. This uses a property known as _variance_.
```

src/subtyping.md
```diff
@@ -1,146 +1,143 @@
...
-how to compose subtyping. Once again, we need *variance*.
+This is a classic use-after-free bug!
 
-## Variance
+Our first instinct might be to blame the `assign` impl, but there's really nothing wrong here.
+It shouldn't be surprising that we might want to assign a `T` into a `T`.
 
-Variance is where things get a bit complicated.
+The problem is that we cannot assume that `&mut &'static str` and `&mut &'b str` are compatible.
+This must mean that `&mut &'static str` should **not** be a *subtype* of `&mut &'b str`,
```
danielhenrymantilla, line 136:
```suggestion
This means that `&mut &'static str`**cannot** be a *subtype* of `&mut &'b str`,
```

src/subtyping.md
danielhenrymantilla, line 139: Could we avoid type constructors and talk of _generic types_ instead? It ought to be clearer. Moreover, it incidentally allows to add the sentence: talking of the variance of a non-generic type makes no sense.

  - We're glossing over `fn(&str) <: fn(&'concrete str)` here, by the way, but I think that's fine

src/subtyping.md
```diff
@@ -152,38 +149,36 @@ A type constructor F's *variance* is how the subtyping of its inputs affects the
...
-* `F` is *covariant* if `F<Sub>` is a subtype of `F<Super>` (subtyping "passes through")
-* `F` is *contravariant* if `F<Super>` is a subtype of `F<Sub>` (subtyping is "inverted")
-* `F` is *invariant* otherwise (no subtyping relationship exists)
+* F is **covariant** if `F<Sub>` is a subtype of `F<Super>` (the subtype property is passed through)
+* F is **contravariant** if `F<Super>` is a subtype of `F<Sub>` (the subtype property is "inverted")
+* F is **invariant** otherwise (no subtyping relationship exists)
 
-If `F` has multiple type parameters, we can talk about the individual variances
-by saying that, for example, `F<T, U>` is covariant over `T` and invariant over `U`.
+If we remember from the above examples,
```
danielhenrymantilla, line 156: FWIW, I agree with @conradludgate's point about the advantage of using `:` everywhere; it allows to use this "subtyping is passed through" / "subtyping is inverted" wave-handed but very useful mnemonic.

I may also be useful to draw a tangential (footnote?) parallel with function monotonicity in mathematics, since for those that know that notion they can suddenly realize they know what all this is all about 🙂

danielhenrymantilla (approved):
I would very much not mix traits into this chapter ⚠️, but other than that, I'm quite fond of this PR 🙂 ✅

src/subtyping.md
jyn514: "degrade"  (like atoms) might be a more descriptive work, or "reduce".

conradludgate:
Thanks everyone for the reviews. It seems overall people are in favour of this general form but there's a bit of bike shedding over syntax.

I'll make all the grammatical and language changes suggested

src/subtyping.md
```diff
@@ -1,146 +1,148 @@
...
-```rust,ignore
-fn love(pet: Animal) {
-    pet.snuggle();
-}
-```
+Subtyping is the idea that one type can be used in place of another.
 
-By default, static types must match *exactly* for a program to compile. As such,
-this code won't compile:
+Let's define that `Sub` is a subtype of `Super` (we'll be using the notation `Sub: Super` throughout this chapter).
```
jackh726, line 48: Just to make sure this doesn't get lost: I don't think we should use this notation. `Sub <: Super` is correct. `Sub: Super` is too confusing with outlives relationship, and generally the opposite of what is (potentially) expected.

conradludgate: Yeah these syntax suggestions are growing on me. I think i agree with them now and I will perform the changes

mejrs: Why not use existing symbols like `⊂` for this, rather than inventing our own?

jackh726: `<:` is well-used to indicate subtyping and is easy to type.

danielhenrymantilla: I think that `<:` seems to be too widespread/common in this scenario not to use it. That being said, when generic types are involved, things start to be quite noise, angle-bracket-wise:

```console
Result<Vec<&'static str>> <: Result<Vec<&'_ str>>
```

So `⊂:`, besides being a nicer smiling face 😄, could be a decent middle-ground with @mejrs's proposal:

```console
Result<Vec<&'static str>> ⊂: Result<Vec<&'_ str>>
```

But, it would have the sad side-effect that it wouldn't be that nice in conjunction with `⊇` for regions, and I _really_ find `⊇` for regions to be an incredibly useful notation:

```rust
for<'lt>
      // 'static ⊇ 'lt, and thus:
    &'static str ⊂ &'lt str
,
```

So, if we were not to use `<:`, I'd rather we use a notation with _motion_ to represent subtyping:

```rust
for<'lt>
      // 'static ⊇ 'lt, and thus:
    &'static str → &'lt str
,
Result<Vec<&'static str>> → Result<Vec<&'_ str>>
```

Something like this is, imho, the most readable notation for people unacquainted with type theory (which, let's not forget it, is the target audience for this kind of chapters).

jackh726:
@conradludgate the types team (wg-traits on zulip) is planning to go over this PR next Friday at 9 am Eastern Time. You're welcome to join if you'd like!

nikomatsakis (commented):
Gave this a read. Some notes:

* I didn't spot any errors, beyond a certain imprecision in distinguishing types/traits/lifetimes and their relationships.
* I've not read the rest of the rustonomicon so I can't really comment on how consistent that usage is, but I think we are definitely going to have to change it eventually. In particular, I think it's we want to have a "sublifetime" relationship (let's write it `<:_l` for now). Even though `'a <:_l 'b iff 'a: 'b`, that doesn't make outlives equivalent, because it's also valid to have an outlives relation between lifetimes and types (and even, in the work I've been doing lately, types and types). 
* That said, I'm not clear 100% on the goals of this book/chapter, so I don't know how much it makes sense to worry at this moment about the distinctions there.
* What did seem to be missing from this chapter was a bit of practical advice: how do you (for example) write a test that variance is what you expect? You don't mention how the compiler infers variance and the fact that it has associated semver/soundness implications, but that seems important.

src/subtyping.md
```diff
@@ -152,38 +154,36 @@ A type constructor F's *variance* is how the subtyping of its inputs affects the
...
+|   | `&'a mut T`     | covariant | invariant         |           |
+|   | `Box<T>`        |           | covariant         |           |
 |   | `Vec<T>`        |           | covariant         |           |
-| * | `UnsafeCell<T>` |           | invariant         |           |
+|   | `UnsafeCell<T>` |           | invariant         |           |
 |   | `Cell<T>`       |           | invariant         |           |
-| * | `fn(T) -> U`    |           | **contra**variant | covariant |
+|   | `fn(T) -> U`    |           | **contra**variant | covariant |
 |   | `*const T`      |           | covariant         |           |
 |   | `*mut T`        |           | invariant         |           |
```
lcnr, line 180: don't need the first row anymore as you're removing the `*`

src/subtyping.md
```diff
@@ -1,189 +1,166 @@
...
+> But this is the Rustonomicon. We're writing unsafe code,
+so we need to understand how this stuff really works, and how we can mess it up.
 
-    // `pet` is an Animal, and Dog is a subtype of Animal,
-    // so this should be fine, right..?
-    *pet = spike;
+Going back to our example above, we can say that `'static <: 'world`.
+For now, let's also accept the idea that subtypes of lifetimes can be passed through references
+(more on this in [Variance](#variance)),
+_e.g._ `&'static str` is a subtype of `&'world str`, then we can let a `&'static str` "downgrade" into a `&'world str`.
```
lcnr, line 75:
```suggestion
_e.g._ `&'static str` is a subtype of `&'world str`, then we can "downgrade" `&'static str` into a `&'world str`.
```

src/subtyping.md
```diff
@@ -1,189 +1,166 @@
...
+With that, the example above will compile:
+
+```rust
+fn debug<T: std::fmt::Debug>(a: T, b: T) {
```
lcnr, line 79:
```suggestion
fn debug<'a>(a: &'a str, b: &'a str) {
```
to stay consistent with the original example

src/subtyping.md
```diff
@@ -315,104 +221,122 @@ All it does is take a mutable reference and a value and overwrite the referent w
...
+Meanwhile, in the caller we pass in `&mut &'static str` and `&'world str`.
 
 Because `&mut T` is invariant over `T`, the compiler concludes it can't apply any subtyping
 to the first argument, and so `T` must be exactly `&'static str`.
 
-The other argument is only an `&'a str`, which *is* covariant over `'a`. So the compiler
-adopts a constraint: `&'spike_str str` must be a subtype of `&'static str` (inclusive),
-which in turn implies `'spike_str` must be a subtype of `'static` (inclusive). Which is to say,
-`'spike_str` must contain `'static`. But only one thing contains `'static` -- `'static` itself!
+This is counter to the `&T` case:
```
lcnr, line 229: there is no `&` here?

Maybe actually stay with `&'a str` instead of using some type parameter `T` here

src/subtyping.md
```diff
@@ -197,116 +174,45 @@ For more types, see the ["Variance" section][variance-table] on the reference.
 > take references with specific lifetimes (as opposed to the usual "any lifetime",
 > which gets into higher rank lifetimes, which work independently of subtyping).
```
lcnr, line 175: that's just wrong, higher rank lifetimes don't work independently of subtyping.

you don't have to fix that here as it's an existing issue

lcnr (commented):
a few nits, then r=me.

there are also some other open reviews you may want to deal with before merging

src/subtyping.md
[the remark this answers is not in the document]
rib: > Nit: subtyping applies to types, not lifetimes

I'm not sure that's really a nitpick. I think it's potentially the crux of why this chapter explaining lifetime variance has been quite unintuitive (at least for me) - since it tries too hard to compare lifetime variance to subtyping which is a different thing.

I've tried to expand on this here: https://github.com/rust-lang/nomicon/issues/339#issuecomment-1422694052 as well as link back to other issues that discussed how rust should define variance for lifetimes.

jyn514:
oof, it's been a while ... what's the current state of this PR?

I would prefer to merge this even if it's not 100% complete rather than have it hang in limbo; we can always improve it in follow-ups and my understand is it's already more accurate than the existing docs.

ehuss (approved):
Thanks @conradludgate for working on this!

I went through and applied some of the review suggestions. If anyone has anything more specific that they would like to see, feel free to open a new PR to update this.

PR #343, opened 2022-02-21
ffi: explicitly declare hello_from_rust for C99

mqy:


src/ffi.md
```diff
@@ -281,6 +281,10 @@ We'll create a C file to call the `hello_from_rust` function and compile it by `
 C file should look like:
 
 ```c
+// Explicitly declare to avoid error like this:
+// implicit declaration of function 'hello_from_rust' is invalid in C99 ...
```
JohnTitor, line 285: I think this comment isn't much helpful as most readers wouldn't wonder why the `extern void hello_from_rust();` line exists, I assume. Could you remove it?

mqy: Removed

JohnTitor (approved):
Thanks!

PR #344, opened 2022-03-17
Add an explanation shared to exclusive transmute

ramnivas:
The current version mentions that "Transmuting an `&` to `&mut` is UB"
and then the sub-bullets (passionately!) dissuades people from doing so,
but without explaining a possible failure scenario. This change alludes
to the fact that such transmutation will run afoul of the assumptions
that the optimizer may legitimately make.


ramnivas:
Context: https://users.rust-lang.org/t/unsafe-code-review-request/73105.

JohnTitor (approved):
Thanks!

PR #346, opened 2022-04-05
Change "writers" to "readers" for Deref.

hkBst:
Deref and DerefMut code comments both use "writers", but Deref should use "readers".


JohnTitor (approved):
Thanks!

PR #347, opened 2022-04-22
Update intro.md

clearydude:
Very dumb PR/fix but type punning -> pinning . Nothing funny about types!


Gankra:
No punning is the right thing. It's "saying some piece of memory has two different types because you know the layouts overlap in interesting ways". Exactly what transmute and friends do. 

I don't think the nomicon even talks about Pin?

Gankra:
(thanks for the PR tho, appreciate typo fixes in general!)

clearydude:
Whoops [TIL](https://en.wikipedia.org/wiki/Type_punning)!

PR #349, opened 2022-05-04
"UB" vs "Undefined Behavior"

Arthur-Milchior:
The first occurrence of "UB" appeared without being ever introduced as an
abbreviation. It took me some time to get what it means. It was not had, but I
suspect that the space it wins in a manual is not worth its use here, so I
suggest to replace it by the two words.

However, I also suggest introduce this abbreviation explicitly, which will
prepare people who join some unsafe Rust discussion outside of this book.


JohnTitor (approved):
Thanks!

PR #351, opened 2022-05-05
Expliciting lifetime in the first example

Arthur-Milchior:
The main issue with having three lets one after the other, is that it's not
clear if the lifetime created due to `let x` is the one before or after this
line. Assuming I understood correctly what's going on, this example will clarify
that the lifetime is created after the let. I also add a note stating what the
usual notation is and warning that this longer explicit lifetime name is used
only once for the sake of being as clear as possible.

Other solutions I considered would be to add a "print" between two lets, this
will ensure that the lifetime introduction is near a single "let", hence making
it clear which "let" is concerned by the lifetime.

If I got it wrong and actually the scope is introduced before the let statement,
I guess at least this confirms that clarification may be helpful.

### DIFF
--- a/src/lifetimes.md
+++ b/src/lifetimes.md
@@ -32,9 +32,10 @@ explicitly is *extremely noisy*. All Rust code relies on aggressive inference
 and elision of "obvious" things.
 
 One particularly interesting piece of sugar is that each `let` statement
-implicitly introduces a scope. For the most part, this doesn't really matter.
-However it does matter for variables that refer to each other. As a simple
-example, let's completely desugar this simple piece of Rust code:
+implicitly introduces a scope immediately after the stateme. For the most part,
+this doesn't really matter.  However it does matter for variables that refer to
+each other. As a simple example, let's completely desugar this simple piece of
+Rust code:
 
 ```rust
 let x = 0;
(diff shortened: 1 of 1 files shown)


JohnTitor (changes requested):
I don't think changing lifetime names on later examples is easy to read, saying like "`'a` implies the global lifetime" should be enough.
Note that the nomicon is less of a book for _newcomers_, that should defer to "the book" or something else.

src/lifetimes.md
```diff
@@ -32,9 +32,10 @@ explicitly is *extremely noisy*. All Rust code relies on aggressive inference
 and elision of "obvious" things.
 
 One particularly interesting piece of sugar is that each `let` statement
-implicitly introduces a scope. For the most part, this doesn't really matter.
-However it does matter for variables that refer to each other. As a simple
-example, let's completely desugar this simple piece of Rust code:
+implicitly introduces a scope immediately after the stateme. For the most part,
```
JohnTitor, line 35:
```suggestion
implicitly introduces a scope immediately after the declaration. For the most part,
```

Arthur-Milchior:
I beg your pardon, does "waiting-on-review" mean you wait for another reviewer, or that you wait for me to do a change?
I was waiting for an answer to my last question before potentially updating this PR

src/lifetimes.md
```diff
@@ -47,21 +48,24 @@ likely desugar to the following:
...
         }
     }
 }
 ```
 
-Wow. That's... awful. Let's all take a moment to thank Rust for making this easier.
+Wow. That's... awful. Let's all take a moment to thank Rust for making this
+easier. To make matter worse, traditionally, lifetimes are not as descriptive
+and simply called `'a`, `'b`, and so on. We will respect this tradition going
+forward.
```
JohnTitor, line 68: I don't think this change is much helpful for readers and worth rephrasing so. As stated in the intro, we assume readers have a basic knowledge of Rust, actually, https://doc.rust-lang.org/nightly/book/ch10-03-lifetime-syntax.html#lifetime-annotation-syntax explains it and that should be enough.

src/lifetimes.md
```diff
@@ -47,21 +48,24 @@ likely desugar to the following:
...
-        let y: &'b i32 = &'b x;
-        'c: {
-            // ditto on 'c
-            let z: &'c &'b i32 = &'c y;
+    'x_lifetime: {
+        // lifetime used is 'x_lifetime because that's good enough.
+        let y: &'x_lifetime i32 = &'x_lifetime x;
+        'y_lifetime: {
+            // ditto on 'y_lifetime
+            let z: &'y_lifetime &'x_lifetime i32 = &'y_lifetime y;
```
JohnTitor, line 59: After some re-thinking of this, I still think this is too descriptive. We easily can find `'a` is a lifetime and curly braces are scopes if we're familiar with Rust's syntax. That is, _naming_ lifetimes looks redundant, and tweaking prior wording should be enough here.

JohnTitor:
> I beg your pardon, does "waiting-on-review" mean you wait for another reviewer, or that you wait for me to do a change? I was waiting for an answer to my last question before potentially updating this PR

It implies a PR is waiting on review.

PR #354, opened 2022-05-05
Clarification of borrowck

Arthur-Milchior:
Same as other recent PR with abbreviation

### DIFF
--- a/src/borrow-splitting.md
+++ b/src/borrow-splitting.md
@@ -1,10 +1,10 @@
 # Splitting Borrows
 
 The mutual exclusion property of mutable references can be very limiting when
-working with a composite structure. The borrow checker understands some basic
-stuff, but will fall over pretty easily. It does understand structs
-sufficiently to know that it's possible to borrow disjoint fields of a struct
-simultaneously. So this works today:
+working with a composite structure. The borrow checker (a.k.a. borrowck)
+understands some basic stuff, but will fall over pretty easily. It does
+understand structs sufficiently to know that it's possible to borrow disjoint
+fields of a struct simultaneously. So this works today:
 
 ```rust
 struct Foo {


src/borrow-splitting.md
```diff
@@ -23,8 +23,8 @@ let c2 = &x.c;
 println!("{} {} {} {}", a, b, c, c2);
 ```
 
-However borrowck doesn't understand arrays or slices in any way, so this doesn't
-work:
+However the borrow checker doesn't understand arrays or slices in any way, so
```
JohnTitor, line 26: Why did you change the term even though we explained `dropck` in the previous section? Note that `dropck` is a well-used term on Rust and it's unnecessary to say "borrow checker" so aggressive.

Arthur-Milchior: Hi,

I assume you meant "borrowck" since this PR don't touch "dropck".

I used the same heuristic than in other PR I did recently that it's nice to avoid abbreviation, unless the texts become significatively longer. I love this rule in general, however I admit it is useless if other documents don't follow it, and in this case introducing it once and then using it to let the reader be used to it makes a ton of sense. 
If we go with the second notion, then I'm fine changing the PR so that it simply adds "(a.k.a. borrowck )", so at least it'll be introduced once.

JohnTitor: > I assume you meant "borrowck" since this PR don't touch "dropck".

Yeah, you're right.


> I used the same heuristic than in other PR I did recently that it's nice to avoid abbreviation, unless the texts become significatively longer. I love this rule in general, however I admit it is useless if other documents don't follow it, and in this case introducing it once and then using it to let the reader be used to it makes a ton of sense. If we go with the second notion, then I'm fine changing the PR so that it simply adds "(a.k.a. borrowck )", so at least it'll be introduced once.

I think not using "borrowck" only on this doc is somewhat weird, so yes, let's just adding to the first one.

Arthur-Milchior:
Thenk I've just added "a.k.a. borrowck" to be explicit on first use, and kept the standard abbreviation every where else.

By the way, thanks a lot for accepting most of my changes and taking time to discuss the ones that are not valid. And obviously thanks for taking care of this very interesting documentation

JohnTitor (approved):
Thanks for making those changes too :)

PR #366, opened 2022-05-29
Fix wording on the aliasing section

### LINKED ISSUE #362: Is the proof of  the assume that`temp` doesn't alias `input`wrong in section 3.2
The doc says:

> We're still relying on alias analysis to assume that temp doesn't alias input, but the proof is much simpler: the value of a local variable can't be aliased by things that existed before it was declared. This is an assumption every language freely makes, and so this version of the function could be optimized the way we want in any language.

The assumption is that `temp` doesn't ALIAS `input`, but the proof is try to prove `input` doesn't ALIAS local variable `temp`.

JohnTitor:
It should've said "the value of a local variable cannot alias
things...".

Fixes #362

### DIFF
--- a/src/aliasing.md
+++ b/src/aliasing.md
@@ -118,8 +118,8 @@ fn compute(input: &u32, output: &mut u32) {
 }
 ```
 
-We're still relying on alias analysis to assume that `temp` doesn't alias
-`input`, but the proof is much simpler: the value of a local variable can't be
+We're still relying on alias analysis to assume that `input` doesn't alias
+`temp`, but the proof is much simpler: the value of a local variable can't be
 aliased by things that existed before it was declared. This is an assumption
 every language freely makes, and so this version of the function could be
 optimized the way we want in any language.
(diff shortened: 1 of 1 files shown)


ehuss:
I've read this over several times, and I'm not clear on this.

The original says: "the value of a local variable can't be aliased by things that existed before it was declared" which I read to be true.  The local variable can't be aliased before it is declared because it doesn't exist, yet. There is nothing to alias. That is:

```rust
fn compute(input: &u32, output: &mut u32) {
    // Up to this point, nothing can alias `temp` because `temp` doesn't exist, yet.
    let mut temp = *output;
    if *input > 10 {
        temp = 1;
    }
    if *input > 5 {
        temp *= 2;
    }
    *output = temp;
}
```

The new version says: "the value of a local variable cannot alias things that existed before it was declared."
This doesn't seem correct to me because local variables can be created which alias things before their declaration.

```rust
fn f(a: &str) {
    let b = a;  // local variable `b` created an alias of `a`
}
```

Can you help me understand the proposed correction?

JohnTitor:
Thanks for pointing out, hmm, I was confused then. So, we should say "We're still relying on alias analysis to assume that `input` doesn't alias `temp`" instead, right?

JohnTitor:
Friendly-ping @ehuss it'd be great if you could take another look :)

ehuss (approved):
I'm going to go ahead and approve, though I'm not sure swapping the terms matters much. I would think that aliasing is symmetric (if foo aliases bar, then bar aliases foo).

PR #372, opened 2022-06-29
Update lifetimes.md

zhenya-1007:
Get rid of the obviously wrong (and potentially confusing) double lifetime annotation.  My guess as to the etiology is an errant copy-and-paste.

### DIFF
--- a/src/lifetimes.md
+++ b/src/lifetimes.md
@@ -55,7 +55,7 @@ likely desugar to the following:
         let y: &'b i32 = &'b x;
         'c: {
             // ditto on 'c
-            let z: &'c i32 = &'c y;
+            let z: &'c &'b i32 = &'c y; // "a reference to a reference to an i32" (with lifetimes annotated)
         }
     }
 }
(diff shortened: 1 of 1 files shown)


ehuss:
Hi, thanks for the PR!

This doesn't look obviously wrong to me, so can you say more about what you think is wrong about it? This is intentionally not Rust syntax, and is intended as an illustration of desugaring. `z` is a double reference to `x`. The inner reference has a lifetime of `'b` and the outer reference has a lifetime of `'c`.  The example seems to be trying to illustrate this.

zhenya-1007:
Thank you for taking a look.

> `z` is a double reference to `x`. The inner reference has a lifetime of `'b` and the outer reference has a lifetime of `'c`.

With this explanation, it makes sense.  Without it, it looked like a copy-and-paste slip-up to me.  I think that's because there are no other examples (that I could find) where lifetime annotations appear twice in a de-sugaring.

ehuss (approved):
Thanks!

PR #374, opened 2022-07-18
NonNull::dangling uses `align_of<T>`, not `size_of<T>`

5225225:
This reverts #371.

Because `NonNull::dangling()` does *not* use the size of T, otherwise it would have no way to make a pointer to a ZST.


Gankra:
thanks

PR #376, opened 2022-07-27
Update the `repr(transparent)` section to reflect the current state

### LINKED ISSUE #375: repr(transparent) works on single variant enums too
The section "[repr(transparent)](https://github.com/rust-lang/nomicon/blob/8d1e4dccf71114ff56f328f671f2026d8e6b62a2/src/other-reprs.md#reprtransparent)" currently doesn't mention that `#[repr(transparent)]` also works with single variant enums, see:

* [RFC 2645](https://rust-lang.github.io/rfcs/2645-transparent-unions.html)
* [PR 68122 in Rust](https://github.com/rust-lang/rust/pull/68122) (stabilization of transparent `enum`s in 1.42.0)

JohnTitor:
Fixes #375

Signed-off-by: Yuki Okushi <jtitor@2k36.org>

### DIFF
--- a/src/other-reprs.md
+++ b/src/other-reprs.md
@@ -56,8 +56,7 @@ compiled as normal.)
 
 ## repr(transparent)
 
-This can only be used on structs with a single non-zero-sized field
-or enums with only one non-zero-sized field (there may be additional zero-sized fields).
+`#[repr(transparent)]` can only be used on a struct or single-variant enum that has a single non-zero-sized field (there may be additional zero-sized fields).
 The effect is that the layout and ABI of the whole struct/enum is guaranteed to be the same as that one field.
 
 > NOTE: There's a `transparent_unions` nightly feature to apply `repr(transparent)` to unions,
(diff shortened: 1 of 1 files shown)


src/other-reprs.md
```diff
@@ -56,24 +56,27 @@ compiled as normal.)
 
 ## repr(transparent)
 
-This can only be used on structs with a single non-zero-sized field (there may
-be additional zero-sized fields). The effect is that the layout and ABI of the
-whole struct is guaranteed to be the same as that one field.
+This can only be used on structs with a single non-zero-sized field
+or enums with only one non-zero-sized field (there may be additional zero-sized fields).
```
ehuss, line 60: I think it might be good to be explicit that the enum can only have a single variant.  I might also rework this a little to avoid the duplication, so maybe something like:

```suggestion
`#[repr(transparent)]` can only be used on a struct or single-variant enum that has a single non-zero-sized field (there may be additional zero-sized fields).
```

ehuss (commented):
Thanks!  Sorry about the long delay.

JohnTitor:
Applied!

ehuss (approved):
Thanks!

PR #378, opened 2022-08-04
Rewrite atomics section

SabrinaJewson:
This PR is for my attempted rewrite of the atomics section of the Nomicon, to give it a more spec-focused explanation and avoid misconceptions like time and reordering. Currently it’s far from finished, containing only an explanation of multi-threaded execution and the start of the “relaxed” section, but I’m opening this PR for early review and feedback on the explanation method.

I also copy-pasted in the C++ spec, but changed a couple things, due things like the lack of `consume` and `sig_atomic_t`:
* I removed the definition of “dependency carrying”
* I removed definition of “depedency-ordered before”
* I removed the definition of “inter-thread happens before”
* I removed the [happens-before cycle restriction](https://eel.is/c++draft/intro.races#10.sentence-2), since [“This cycle would otherwise be possible only through the use of consume operations.”](https://eel.is/c++draft/intro.races#note-9)
* I replaced the definition of “happens before” with the definition of “simply happens before” (since [“In the absence of consume operations, the happens before and simply happens before relations are identical”](https://eel.is/c++draft/intro.races#note-10))
* I removed the [`volatile sig_atomic_t` exception](https://eel.is/c++draft/intro.races#22)
* I rewrote `memory_order` as `core::sync::atomic::Ordering`
* I removed `kill_dependency`
* I renamed `atomic_thread_fence` to `fence` and `atomic_signal_fence` to `compiler_fence`, as it is called in std

I’m not certain about the copyright of this — ISO says that they only hold copyright over published versions, but I couldn’t find information about copyright of drafts.


src/atomics/relaxed.md
```diff
@@ -0,0 +1,43 @@
+# Relaxed
+
+Now we’ve got single-threaded mutation semantics out of the way, we can try
+reintroducing a second thread. We’ll have one thread perform a write to the
+memory location, and a second thread read from it, like so:
+
+```rs
+// Initial state
+let mut state = 0;
```
ibraheemdev, line 9:
```suggestion
let mut data = 0;
```

src/atomics/relaxed.md
```diff
@@ -0,0 +1,43 @@
...
+// Thread 1:
+data = 1;
+// Thread 2:
+data;
```
ibraheemdev, line 13:
```suggestion
println!("{data}");
```

Makes it more clear this is a read.

jwakely:
> I’m not certain about the copyright of this — ISO says that they only hold copyright over published versions, but I couldn’t find information about copyright of drafts.

https://en.cppreference.com/w/cpp/language/memory_model and https://en.cppreference.com/w/cpp/atomic/memory_order#Formal_description paraphrase the relevant parts of the C++ spec, and are CC-BY-SA and GFDL dual-licensed. See [What can I do with the material on this site?](https://en.cppreference.com/w/Cppreference:FAQ#What_can_I_do_with_the_material_on_this_site.3F).

src/atomics/multithread.md
```diff
@@ -0,0 +1,220 @@
...
+The first important thing to understand about C++20 atomics is that **the
+abstract machine has no concept of time**. You might expect there to be a single
+global ordering of events across the program where each happens at the same time
+or one after the other, but under the abstract model no such ordering exists;
+instead, a possible execution of the program must be treated as a single event
+that happens instantaneously — there is never any such thing as “now”, or a
+“latest value”, and using that terminology will only lead you to more confusion.
+(Of course, in reality there does exist a concept of time, but you must keep in
+mind that you’re not programming for the hardware, you’re programming for the
+AM.)
```
Cassy343, line 12: This acronym is never defined, maybe add "(AM)" the first time you mention "abstract machine/model," or just type it in full here.

src/atomics/multithread.md
```diff
@@ -0,0 +1,220 @@
+# Multithreaded Execution
+
+The first important thing to understand about C++20 atomics is that **the
+abstract machine has no concept of time**. You might expect there to be a single
```
Cassy343, line 4: Good to front it with this, but it might also be worth a couple small paragraphs differentiating the abstract machine from the physical machine, and perhaps motivate why we care to make that distinction.

src/atomics/multithread.md
```diff
@@ -0,0 +1,220 @@
...
+╰───╥───╯     ?╌┼╌╌╌╌┤
+╭───⇓───╮     ?╌┼╌╌╌╌┤
+│  = 2  ├╌?     │  ? │
+╰───────╯       └────┘
+```
+
+Note the use of dashed padding in between the values of `data`’s column. Those
+spaces won’t ever contain a value, but they’re used to represent an
+unsynchronized (non-atomic) write — it is garbage data and attempting to read it
+would result in a data race.
```
Cassy343, line 135: It would've helped expedite my reading if you gave a brief example of what a write looked like (ie connecting to two of the dashed lines), and that a read connected to the middle. I only fully understood what you meant here after reading down to the completed diagram. I think this visual is great but could've just been fronted with a small example/key.

src/atomics/multithread.md
```diff
@@ -0,0 +1,220 @@
...
+╭───⇓───╮  ├╌╌┼╌╌╌╌┤
+│  = 2  ├╌╌┘  │  2 │
+╰───────╯     └────┘
+```
+
+You might be thinking that all this has been is the longest, most convoluted
+explanation ever of the most basic intuitive semantics of programming — and
+you’d be absolutely right. But it’s essential to grasp these fundamentals,
+because once you have this model in mind, the extension into multiple threads
+and the complicated semantics of real atomics becomes completely natural.
```
Cassy343, line 220: Love this intro! I don't know if this is finalized or anything but showing what a data race looks like after this might be helpful.

Cassy343: I jumped the gun on this comment - definitely reasonable that you introduce that example in the relaxed section.

src/atomics/relaxed.md
```diff
@@ -0,0 +1,423 @@
...
+```
+
+Note that **both sides must be atomic to avoid the data race**: if only the
+writing side used atomic operations, the reading side would still have multiple
+values to choose from (UB), and if only the reading side used atomic operations
+it could end up reading the garbage data “in-between” `0` and `1` (also UB).
+
+> **NOTE:** This description of why both sides are needed to be atomic
+> operations, while neat and intuitive, is not strictly correct: in reality the
+> answer is simply “because the spec says so”. However, it is isomorphic to the
```
Cassy343, line 104: Maybe substitute "isomorphic" for "equivalent"

src/atomics/relaxed.md
```diff
@@ -0,0 +1,423 @@
...
+
+```rust
+# use std::sync::atomic::{self, AtomicU64};
+static COUNTER: AtomicU64 = AtomicU64::new(0);
+pub fn get_id() -> u64 {
+    COUNTER.fetch_add(1, atomic::Ordering::Relaxed)
+}
+```
+
+And then, no matter how many threads there are, that race condition from earlier
```
Cassy343, line 164: Was the distinction between a race condition and a data race ever specified? I may have gotten my words mixed up there but IIRC one is just unspecified but defined behavior, but the other is UB.

src/atomics/relaxed.md
```diff
@@ -0,0 +1,423 @@
...
+Thread 1   COUNTER   Thread 2
+╭───────╮   ┌───┐   ╭───────╮
+│ load  ├───┤ 0 ├───┤  load │
+╰───╥───╯   └───┘   ╰────╥──╯
+╭───⇓───╮ ┌─┬───┐   ╭────⇓──╮
+│ store ├─┘ │ 1 │ ┌─┤ store │
+╰───────╯   └───┘ │ ╰───────╯
+            ┌───┬─┘
+            │ 1 │
+            └───┘
```
Cassy343, line 137: Maybe leave a note that the fact that `1` appears twice sequentially in the MO is intentional

src/atomics/relaxed.md
```diff
@@ -0,0 +1,423 @@
...
+`compare_exchange` is a bit like a store, but instead of unconditionally storing
+the value, it will first check the previous value in the modification order to
+see whether it is what we expect, and if not it will simply tell us that and not
+make any changes. It is an RMW operation, so all of this happens fully
+atomically — there is no chance for a race condition.
+
+> \* It’s not quite the same, because `compare_exchange` can suffer from ABA
+> problems in which it will see a later value in the modification order that
+> just happened to be same and succeed. However, in this code values can never
+> be reused so we don’t have to worry about that.
```
Cassy343, line 218: This is the first time "ABA problem" is mentioned and it might be helpful to put some concrete values into a one-sentence example. Like "for instance, the modification order could contain the sub-sequence (1, 2, 1), and while the `compare_exchange` was intended to be performed on the first 1, it would also succeed on the second."

src/atomics/relaxed.md
```diff
@@ -0,0 +1,423 @@
...
+load in the atomic’s modification order, and that leads to the races. What we
+need is some way to say, “add this new value to the modification order, but
+_only if_ it occurs directly after the value we loaded”. And luckily for us,
+there exists a function that does exactly\* this: `compare_exchange`.
+
+`compare_exchange` is a bit like a store, but instead of unconditionally storing
+the value, it will first check the previous value in the modification order to
+see whether it is what we expect, and if not it will simply tell us that and not
+make any changes. It is an RMW operation, so all of this happens fully
+atomically — there is no chance for a race condition.
```
Cassy343, line 213: So is this saying that a RMW always loads the last value in the MO?

src/atomics/acquire-release.md
```diff
@@ -0,0 +1,334 @@
...
+weaker than the _sequenced before_ double-arrows (⇒) that occur inside a single
+thread, but can still be used with the arrow rules to determine which values of
+a memory location are valid to read. We can say that in the first possible
+execution, Thread 1’s `store` (and everything sequenced before that) _happens
+before_ Thread 2’s load (and everything sequenced after that).
+
+There is one more rule required for these to be useful, and that is _release
+sequences_: after a release store is performed on an atomic, happens before
+arrows will connect together each subsequent value of the atomic as long as the
+new value is caused by an RMW and not just a plain store.
```
Cassy343, line 209: Don't release stores participate in the release sequence as well or is this a gap in my understanding?

src/atomics/acquire-release.md
```diff
@@ -0,0 +1,334 @@
...
+╭───⇓───╮ └╌│╌╌╌╌╌╌╌╌╌╌╌╌┘         ╭───⇓───╮
+│ store ├─┐ │ ┌───────↙────────────┤ store │
+╰───────╯ │ │ │ false │            ╰───────╯
+          │ │ └───┬───┘
+          │ └─┬───↓───┐
+          │   │ true  │
+          │   └───────┘
+          └───↘───────┐
+              │ false │
+              └───────┘
```
Cassy343, line 291: Maybe connect the `guard` box in T2 for sake of completeness

src/atomics/acquire-release.md
SabrinaJewson: Until C++20, stores performed by the original thread participated in release sequences but I don’t think arbitrary release stores ever have. [cppreference says this](https://en.cppreference.com/w/cpp/atomic/memory_order#Release_sequence):

> After a release operation A is performed on an atomic object M, the longest continuous subsequence of the modification order of M that consists of
> 1) Writes performed by the same thread that performed A (until C++20)
> 2) Atomic read-modify-write operations made to M by any thread
> is known as _release sequence headed by A_

Cassy343: Interesting. Might be worth calling that caveat out explicitly then.

src/atomics/relaxed.md
jwakely: That's right

src/atomics/multithread.md
SabrinaJewson: I’ve introduced AM as an acronym in the intro (`atomics.md`), which should be enough.

src/atomics/relaxed.md
SabrinaJewson: No, it’s just saying that it loads the previous value. I improved the wording a bit to clarify this and added a paragraph explaining that the “RMWs see the latest value” thing doesn’t make any sense.

SabrinaJewson:
Given that my free time has been reduced since I’ve started this PR and the signals section (which I was part way through writing)  is a mostly different area of expertise to the rest of atomics, I'm going to mark this as ready to review now.

Open questions:
* Are Unicode diagrams the best format? Do they display correctly on all devices? Would it even be possible to use a different format?
* CI is red because Rust’s std docs now link to the broken link `atomics.html` (which is now `atomics/atomics.html`) — what can be done about this?

cc @cbeuw for fact-checking since you seem to be an expert on the model (only if you have time of course).

SabrinaJewson:
I have now [rendered this PR](https://sabrinajewson.org/rust-nomicon/atomics/atomics.html) for easier review.

cbeuw:
This looks absolutely amazing! The Unicode diagrams are nicely rendered on both my Mac and Android phone, and are very pretty. This may well be the best tutorial on C++ memory model out there, including all the ones I've seen targeting C++ exclusively.

I definitely wouldn't call me a weak memory expert, considering whenever I work with it I always come up with new questions instead of being able to tell the definite behaviours (but this may be a good thing sometimes). Unfortunately for me, the next 3 months will be extremely busy. I won't be able to go through it in detail, but I'll have my eye on it from time to time, both now and after it's been merged. There will almost certainly be _extremely_ subtle errors given the topic's nature, but that shouldn't prevent this from going live.

src/atomics/seqcst.md
```diff
@@ -0,0 +1,445 @@
...
+              │ true  ├──┘╭────⇓────╮┌─┘╭────⇓────╮└──┤ true  │
+              └───────┘   │ load Y  ├┘└─┤ load X  │   └───────┘
+                          ╰─────────╯   ╰─────────╯
+```
+
+Which results in a failed assert. This execution is brought about because the
+model of separate modification orders means that there is no relative ordering
+between `X` and `Y` being changed, and so each thread is allowed to “see” either
+order. However, some algorithms will require a globally agreed-upon ordering,
+and this is where `SeqCst` can come in useful.
```
cbeuw, line 88: I'd be interested to know which algorithms actually require SeqCst. The leading motivation for proposing algorithms under SeqCst is probably that it's easier to reason about, not that a Single Total Order is necessary (read-latest may be needed but this can be done with idempotent RMW).

src/atomics/atomics.md
```diff
@@ -0,0 +1,121 @@
+# Atomics
+
+Rust pretty blatantly just inherits the memory model for atomics from C++20. This is not
```
cbeuw, line 3: It's probably good to mention the correspondence to `atomic_ref` again, as was added to the std docs https://doc.rust-lang.org/std/sync/atomic/

src/atomics/seqcst.md
```diff
@@ -0,0 +1,445 @@
...
+As I’ve been alluding to for a while, I wasn’t being totally truthful when I
+said that _S_ is consistent with happens-before relations — in reality, it is
+only consistent with _strongly happens-before_ relations, which presents a
+subtly-defined subset of happens-before relations. In particular, it excludes
+two situations:
+
+1. The `SeqCst` operation A synchronizes-with an `Acquire` or `AcqRel` operation
+   B which is sequenced-before another `SeqCst` operation C. Here, despite the
+   fact that A happens-before C, A does not _strongly_ happen-before C and so is
+   there not guaranteed to precede C in _S_.
```
peterjoel, line 282:
```suggestion
   not guaranteed to precede C in _S_.
```

src/atomics/multithread.md
```diff
@@ -0,0 +1,291 @@
...
+	means it is allowed to result in executions no sane computer would actually
+	generate in real life. It is also purposefully built with forward
+	compatibility in mind, giving compilers the opportunity to make better and
+	more aggressive optimizations in the future. As a result, it can be quite
+	hard to test code, especially if you’re on a system that exploits fewer of
+	the AM’s allowed semantics, so it is highly recommended to utilize tools
+	that intentionally produce these executions like [Loom] and [Miri].
+1. Its model is highly formalized and not representative of what goes on
+	underneath. Because C++ needs to be defined by a formal specification and
+	not just hand-wavy rules about “this is what allowed and this is what
```
peterjoel, line 27:
```suggestion
	not just hand-wavy rules about “this is what is allowed and this is what
```

src/atomics/acquire-release.md
```diff
@@ -0,0 +1,354 @@
...
+        Self {
+            locked: AtomicBool::new(false),
+            data: UnsafeCell::new(data),
+        }
+    }
+}
+```
+
+Now for the lock function. We need to use an RMW here, since we need to both
+check whether it is locked and lock it if is not in a single atomic step; this
```
timvermeulen, line 29: > and lock it **if is not** in a single atomic step

A word is missing here.

src/atomics/acquire-release.md
```diff
@@ -0,0 +1,354 @@
...
+            │ false │
+            └───────┘
+```
+
+We now can trace back along the reverse direction of arrows from the `guard`
+bubble to the `+= 1` bubble; we have established that Thread 2’s load
+happens-after the `+= 1` side effect, because Thread 2’s CAS synchronizes-with
+Thread 1’s store. This both avoids the data race and gives the guarantee that
+`1` will be always read by Thread 2 (as long as locks after Thread 1, of
+course).
```
timvermeulen, line 260: > as long as locks after Thread 1,

A word is missing here, I think?

src/atomics/atomics.md
```diff
@@ -0,0 +1,121 @@
...
+
+However there's a third potential state that the hardware enables:
+
+* `y = 2`: (thread 2 saw `x = 1`, but not `y = 3`, and then overwrote `y = 3`)
+
+It's worth noting that different kinds of CPU provide different guarantees. It
+is common to separate hardware into two categories: strongly-ordered and weakly-ordered.
+Most notably x86/64 provides strong ordering guarantees, while ARM
+provides weak ordering guarantees. This has two consequences for concurrent
+programming:
```
tgross35, line 109: "Strongly ordered" and "weakly ordered" hardware are introduced here and not (I don't think) explained anywhere else. 

Maybe add super short context like:

> It is common to separate hardware into two categories: strongly-ordered (every concurrent task sees memory changes in the order they happen) and weakly-ordered or relaxed (no guarantees for order of changes among tasks, only within tasks). Most notably x86/64 provides strong ordering guarantees, while ARM
provides weak ordering guarantees. This has two consequences for concurrent
programming:

newpavlov:
I have minor rendering issues with the diagrams on Linux (in both Firefox and Chromium):



I suggest replacing with inline SVG

JanBeh:
From [Issue #104814 on Rust](https://github.com/rust-lang/rust/issues/104814):

> […] it should be clarified whether the difference between the Rust documentation and the C++20 reference exists intentionally, i.e. is Rust deliberately giving less guarantees to the programmer than the C++20 standard does? (Even though still following the C++20 memory model in practice "as of now".)

I [have been told](https://users.rust-lang.org/t/better-understanding-atomics/84694/9?u=jbe) that the C++20 reference takes precedence over Rust's documentation, but wanted to raise this issue nonetheless as it may be relevant for future backward compatibility.

WaffleLapkin:
What is the status here? 👀

src/atomics/atomics.md
```diff
@@ -0,0 +1,121 @@
...
+that events that occur in some order on *one* thread, occur in the same
+order on *another* thread. To guarantee this, we must issue special instructions
+to the CPU telling it to be a bit less smart.
+
+For instance, say we convince the compiler to emit this logic:
+
+```text
+initial state: x = 0, y = 1
+
+THREAD 1        THREAD2
```
evaporei, line 90: Silly thing, now on `master` the `THREAD2` has a space in between `THREAD` and `2`.

src/atomics/multithread.md
```diff
@@ -0,0 +1,291 @@
...
+│  data ├╌?   │  1 │
+╰───╥───╯  ┌╌╌┼╌╌╌╌┤
+╭───⇓───╮  ├╌╌┼╌╌╌╌┤
+│  = 2  ├╌╌┘  │  2 │
+╰───────╯     └────┘
+```
+
+Now to sort out the read operation in the middle. We can use the same rule as
+before to trace up to the first write and rule out us reading either the `0`
+value or the garbage that exists between it and `1`, but how to we choose
```
evaporei, line 250: Typo `to` -> `do`

chorman0773 (approved):
All of the information here looks to be correct.

WaffleLapkin:
@ehuss do you still have too much in the review queue?

SabrinaJewson:
I just pushed a commit to fix the incorrect claim that one needs 4 threads to observe the difference between `SeqCst` and `AcqRel`, which was a misunderstanding I had. I also changed the example that shows the need for `SeqCst` to a simpler one that only uses two threads.

The changed sections include [the `SeqCst` chapter](https://sabrinajewson.org/rust-nomicon/atomics/seqcst.html) and [the part on `SeqCst` fences](https://sabrinajewson.org/rust-nomicon/atomics/fences.html#seqcst-fences). Thanks @ibraheemdev for pointing this out!

Noratrieb:
@ehuss it seems like people with expertise in atomics have given this a look by now. so I think it would just need an editorial review from you, which is hopefully quicker than a technical one (it takes some time still of course, and I don't expect you to take a look at this very soon, but just FYI)

traviscross:
We discussed this in the lang-docs call today.  Thanks to everyone here for the reviews and the feedback, and to @SabrinaJewson for this work.  We're trying to have a look at this.  Even just editorially, it's an impressive body of work and may take some time.  Thanks for bearing with us.

src/atomics/fences.md
```diff
@@ -0,0 +1,257 @@
+# Fences
+
+As well as loads, stores, and RMWs, there is one more kind of atomic operation
+to be aware of: fences. Fences can be triggered by the
+[`core::sync::atomic::fence`] function, which accepts a single ordering
+parameter and returns nothing. They don’t do anything on their own, but can be
+thought of as events that strengthen the ordering of nearby atomic operations.
```
ppershing: On first read, I thought "ordering" as of "re-ordering" but now I understand that it means memory ordering. I think it would be very wise to make sure readers understand the difference and cannot confuse the two. I would propose to use ordering in a sense memory-ordering always as `ordering`

src/atomics/seqcst.md
```diff
@@ -0,0 +1,432 @@
...
+and this is where `SeqCst` can come in useful.
+
+This ordering, first and foremost, inherits the guarantees from all the other
+orderings — it is an acquire operation for loads, a release operation for stores
+and an acquire-release operation for RMWs. In addition to this, it gives some
+guarantees unique to `SeqCst` about what values it is allowed to load. Note that
+these guarantees are not about preventing data races: unless you have some
+unrelated code that triggers a data race given an unexpected condition, using
+`SeqCst` can only prevent you from race conditions because its guarantees only
+apply to other `SeqCst` operations rather than all data accesses.
```
ppershing: I wonder whether it would be good to add at this point a note that despite popular belief that seqcst is the "strongest" ordering, it doesn't mean that every SeqCst operation is also Acquire or Release operation - SeqCst load cannot provide Release guarantee and vice versa.

src/atomics/relaxed.md
```diff
@@ -0,0 +1,452 @@
...
+would be totally valid thing to happen. But luckily, the authors of the C++
+specification have recognized this as a problem, and as such refined the
+semantics of `Relaxed` to implement a thorough, logically sound, mathematically
+proven formal model that prevents it, that’s just too complex and technical to
+explain here—
+
+> No “out-of-thin-air” values can be computed that circularly depend on their
+> own computations.
+
+Just kidding. Turns out, it’s a *really* difficult problem to solve, and to my
```
ppershing: It would be nice to link to https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3710.html from "difficult"

src/atomics/fences.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Acquire fences
+
+The most common kind of fence is an _acquire fence_, which can be triggered in
+three different ways:
+1. `atomic::fence(atomic::Ordering::Acquire)`
+1. `atomic::fence(atomic::Ordering::AcqRel)`
+1. `atomic::fence(atomic::Ordering::SeqCst)`
+
+An acquire fence retroactively makes every single non-`Acquire` operation that
```
ppershing: This is subtly wrong. fence will not turn relaxed store into an Acquire operation. Also, fence will not mirucalously turn non-atomic read to Acquire load. I think it would be better to re-phrase this.
I also found "in other words, it causes every prior `Release`d value that was" hard to comprehend. 
I tried to come up with a better wording but I am not sure it is 100% clear yet:
"in other words, for any (even relaxed) atomic load sequenced before the fence, if the load reads a value that was `Released`d by some operation O, it creates synchronizes-with relationship between the fence and O.

PR #386, opened 2022-11-14
Grammar change for 3.4: Limits of Lifetimes (lifetime-mismatch.md)

jamesreprise:
"[...] stops to be needed [...]" -> "[...] ceases to be needed [...]"


ehuss (approved):
Thanks!

I think "stops" and "ceases" are near synonyms. But I think in this case the "ceases" sounds better to me, and I think is a more common idiom in this form.

PR #388, opened 2022-12-17
Discourage the use of transmute to construct invalid values

Niki4tap:
Creating null values of non-null types with transmute is undefined behavior, and the nomicon doesn't explicitly say that in the ["nullable pointer optimization"](https://doc.rust-lang.org/nomicon/ffi.html#the-nullable-pointer-optimization) chapter. Hopefully this comment will discourage the use of transmute in this case a little bit more.


ehuss (approved):
Thanks!

PR #389, opened 2022-12-22
FFI: make gcc command clearer

kraktus:
Didn't check if `gcc` was accepting the flag without space but it definitely reduced the comprehension when first encountering it.

### DIFF
--- a/src/ffi.md
+++ b/src/ffi.md
@@ -293,7 +293,7 @@ We name the file as `call_rust.c` and place it on the crate root.
 Run the following to compile:
 
 ```sh
-gcc call_rust.c -o call_rust -lrust_from_c -L./target/debug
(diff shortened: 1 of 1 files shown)


ehuss:
gcc explicitly recommends not using separate arguments. I'd be a little leery of using a form that isn't recommended (and pretty rare in my experience). I understand that it can look a little strange, but I think I would prefer to keep with established norms.

kraktus:
I see, I was not aware of this convention. it is only established for the `-l` flag? Since other args are separate arguments (well only the output one)

ehuss:
`-l` and `-L` are the ones I am most aware of that usually do not use a space. Long flags usually use an equals or a space.

kraktus:
Ok thanks for the feedback!

PR #391, opened 2023-01-01
borrow-splitting: Use `take` instead of `replace`

xy2iii:
Clippy complains with this lint on the old code: https://rust-lang.github.io/rust-clippy/master/index.html#mem_replace_with_default


JohnTitor (approved):
Thanks!

PR #395, opened 2023-01-08
Fix typo in 3.8 Subtyping and Variance

masloyet:
The sentence refers to lifetimes being covariant, when what I believe is meant is that references are covariant over lifetimes. It might be more correct to use "references to be covariant over lifetimes," but I followed the same structure as "Box (...) to be covariant" earlier in the sentence.


JohnTitor (approved):
Good catch, thank you!

PR #365, opened 2022-05-28
Changes for `c_unwind`

BatmanAoD:
One of the documentation changes for [RFC #2945](https://github.com/rust-lang/rfcs/blob/master/text/2945-c-unwind-abi.md).

PR #396, opened 2023-01-14
Copyedits to FFI and unwinding section

tmandry:
These are minor changes that make the text flow better to me.

cc #365

### DIFF
--- a/src/ffi.md
+++ b/src/ffi.md
@@ -721,9 +721,9 @@ No `transmute` required!
 
 ## FFI and unwinding
 
-It’s important to be mindful of unwinding when working with FFI. Each
-ABI comes in two variants, one with `-unwind` suffix and one without (except
-the `Rust` ABI, which always permits unwinding).
+It’s important to be mindful of unwinding when working with FFI. Most
+ABI strings come in two variants, one with an `-unwind` suffix and one without.
+The `Rust` ABI always permits unwinding, so there is no `Rust-unwind` ABI.
 
 If you expect Rust `panic`s or foreign (e.g. C++) exceptions to cross an FFI
 boundary, that boundary must use the appropriate `-unwind` ABI string.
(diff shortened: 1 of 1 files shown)


src/ffi.md
```diff
@@ -722,16 +722,19 @@ No `transmute` required!
 ## FFI and unwinding
 
 It’s important to be mindful of unwinding when working with FFI. Each
-non-`Rust` ABI comes in two variants, one with `-unwind` suffix and one without. If
-you expect Rust `panic`s or foreign (e.g. C++) exceptions to cross an FFI
-boundary, that boundary must use the appropriate `-unwind` ABI string (note
-that compiling with `panic=abort` will still cause `panic!` to immediately
-abort the process, regardless of which ABI is specified by the function that
-`panic`s).
+ABI comes in two variants, one with `-unwind` suffix and one without (except
```
tmandry, line 725: Wait, I'm not sure this is always true. Surely there must be some ABIs without a well-defined unwind behavior?

cc @BatmanAoD

BatmanAoD: This bit is talking about `panic!` the macro, not actual stack unwinding. It always aborts when the panic mode is `abort`.

(unless I am misunderstanding the question...)

tmandry: Sorry, I'm referring to the idea that every non-Rust ABI has an -unwind variant. It seems totally reasonable for some ABIs not to have a native unwinding mechanism.

ehuss: There are a few listed [here](https://github.com/rust-lang/rust/blob/4b51adf6ffa1ae2286ea436eb48cbaa8771886e6/compiler/rustc_target/src/spec/abi.rs#L94-L109) which don't have `-unwind` variants.  All of them are unstable/experimental. I would just change it from "Each ABI" to "Most ABIs".

BatmanAoD: That makes sense, and I'll make the change; though I'm not aware of any reason inherent to the ABI definitions themselves that would make `unwind` inappropriate.

BatmanAoD:
Who needs to review this?

PR #398, opened 2023-01-27
Include elided lifetime in sub-typing example

sgasse:
One step required to understand the argumenation is that the lifetime of the `&mut T` is different from any lifetime that may be part of a reference forming the type T. By mentioning the elided lifetime explictly, the paragraph should be easier to understand.

Minor other changes are highlighting types with backticks.

### DIFF
--- a/src/subtyping.md
+++ b/src/subtyping.md
@@ -187,6 +187,10 @@ some sense "fundamental". All the others can be understood by analogy to the oth
 * `*const T` follows the logic of `&T`
 * `*mut T` follows the logic of `&mut T` (or `UnsafeCell<T>`)
 
+Note that type `T` itself can be a reference to another type. So the type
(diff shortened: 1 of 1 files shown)


sgasse:
@Gankra I think you were the last person to thoroughly (re)write the sub-typing part. Thanks a lot for it, I really like the examples with the meowing dog :grin: 

Would you have time to take a quick glance at the proposed change?

MPvHarmelen:
My 2ct as a first-time reader of these sections:

I don't think adding explicit lifetimes is an improvement here: the elided lifetimes weren't necessary to understand what was going on for me. Introducing explicit lifetimes, however, would add quite a bit of visual and textual clutter, distracting me from the lifetimes that matter most.

JohnTitor:
Thanks for the PR! I, however, agree with @MPvHarmelen, this doesn't seem to improve its meaning much. Closing as rejected.

PR #401, opened 2023-02-12
Fixes double bug in Send-Sync example

Mottl:
Fixes double bug in the call to libc::posix_memalign which causes SIGSEGV due to incorrect pointer parameter and EINVAL (retval=22) due to incorrect alignment when size of T is less then size of usize.


JohnTitor (approved):
LGTM, thank you!

PR #403, opened 2023-03-10
Correct possible UB in Arc::drop

krtab:
Hi!

I'm not entirely sure but I believe the code as currently presented as potential for UB. Indeed, `inner` is dropped after the anonymous Box, which means that is is dangling at some point. Iiuc, dangling references are UB in Rust, even if not accessed. In case there is no issue, I think it would be beneficial to explain why in the code comments.


src/arc-mutex/arc-drop.md
```diff
@@ -89,9 +89,13 @@ Now, let's wrap this all up inside the `Drop` implementation:
...
-        let inner = unsafe { self.ptr.as_ref() };
-        if inner.rc.fetch_sub(1, Ordering::Release) != 1 {
-            return;
+        // We create a scope so that inner won't be live
+        // when we create a Box to drop the `ArcInner`
+        {
+            let inner = unsafe { self.ptr.as_ref() };
+            if inner.rc.fetch_sub(1, Ordering::Release) != 1 {
+                return;
+            }
```
krtab, line 98: Hm, if this is indeed not safe I believe that my PR does not completely solve the problem. Another thread can intersperse itself after `rc` has been decreased do the whole drop glue before we leave the scope. In this case, I think that any reference should be avoided.

chorman0773 (commented):
Reading the code, I don't think this change is necessary - the prior code did not have undefined behaviour after the change to `&impl !Freeze`.

A more comprehensive rationale is posted on zulip.

PR #404, opened 2023-03-12
Change incorrect reference to "above" in ffi.md

majaha:
This sentence says "above", but the linked section is actually below.


ehuss (approved):
Thanks!

PR #408, opened 2023-05-09
Expand unbounded lifetime example code and improve wording

Enselic:
Due to wrapping, the diff looks larger than it is. I have only changed

> The most common source of this is dereferencing a raw pointer, which produces a reference with an unbounded lifetime.

into the more accurate

> The most common source of this is taking a reference to a dereferenced raw pointer, which produces a reference with an unbounded lifetime.

See [Zulip ](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Ref.20from.20raw.20pointer.20deref.3F) discussion.

I have also expanded on the example code by expanding:

> For instance:
> ```rust,ignore
> fn get_str<'a>() -> &'a str;
> ```
> will produce an `&str` with an unbounded lifetime.

into:

> For instance:
> 
> ```rust
> fn get_str<'a>(s: *const String) -> &'a str {
>     unsafe { &*s }
> }
> 
> fn main() {
>     let soon_dropped = String::from("hello");
>     let dangling = get_str(&soon_dropped);
>     drop(soon_dropped);
>     println!("Invalid str: {}", dangling); // Invalid str: gӚ_`
> }
> ```

### DIFF
--- a/src/unbounded-lifetimes.md
+++ b/src/unbounded-lifetimes.md
@@ -17,7 +17,8 @@ boundaries.
 Given a function, any output lifetimes that don't derive from inputs are
 unbounded. For instance:
 
-```rust
+<!-- no_run: This example exhibits undefined behavior. -->
+```rust,no_run
 fn get_str<'a>(s: *const String) -> &'a str {
     unsafe { &*s }
 }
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks! I pushed a small change to mark the example as `no_run` since running code with UB in the tests could cause problems.

PR #410, opened 2023-05-14
Update to 2021 edition

LucasFA:
Is there any reason it is 2018 instead of 2021? Everything passed `mdbook test`


ehuss (approved):
Thanks! I don't think there was any particular reason.

PR #411, opened 2023-05-30
phantom-data: Add `Send` and `Sync` columns

### LINKED ISSUE #320: PhantomData patterns table should include interactions with `Send`/`Sync`
According to the `PhantomData` patterns table, there seems to be no difference between using e.g. `*const T` and `fn() -> T`. Both are covariant on `T` and have no lifetime. But there actually is a difference, which is that `PhantomData<*const T>` is not `Send` or `Sync`, whereas `PhantomData<fn() -> T>` is `Send` and `Sync`.

This is something that users can figure out by thinking about how `PhantomData` implements auto traits according to its `T` parameter, but this isn't explicitly called out in the nomicon (or in the `std::marker::PhantomData` docs) and so it's easy to forget. For example, I changed some code from `PhantomData<T>` to `PhantomData<fn() -> T>` specifically to avoid the drop check (my type doesn't own a `T`, but instead it can produce `T`s). It just so happens that this makes my type `Send` and `Sync` too, which it should have been in the first place but nobody realized that until well after the change.

This could be improved by updating the Nomicon to call out the fact that `PhantomData` will implement auto traits according to its type parameter, and to give an example of this by including `Send`/`Sync` columns in the table that lists whether the given pattern is `Send`/`Sync`. This would look something like

**Phantom type** | `'a` | `T` | `Send` | `Sync`
-|-|-|-|-
`PhantomData<T>` | - | covariant (with drop check) | `T: Send` | `T: Sync`
`PhantomData<&'a T>` | covariant | covariant | `T: Sync` | `T: Sync`
`PhantomData<&'a mut T>` | covariant | invariant | `T: Send` | `T: Sync`
`PhantomData<*const T>` | - | covariant | - | -
`PhantomData<*mut T>` | - | invariant | - | -
`PhantomData<fn(T)>` | - | contravariant | `Send` | `Sync`
`PhantomData<fn() -> T>` | - | covariant | `Send` | `Sync`
...

Enselic:
Closes #320

### DIFF
--- a/src/phantom-data.md
+++ b/src/phantom-data.md
@@ -24,7 +24,7 @@ We do this using `PhantomData`, which is a special marker type. `PhantomData`
 consumes no space, but simulates a field of the given type for the purpose of
 static analysis. This was deemed to be less error-prone than explicitly telling
 the type-system the kind of variance that you want, while also providing other
-useful things such as the information needed by drop check.
+useful things such as auto traits and the information needed by drop check.
 
 Iter logically contains a bunch of `&'a T`s, so this is exactly what we tell
 the `PhantomData` to simulate:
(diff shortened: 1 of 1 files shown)


src/phantom-data.md
```diff
@@ -234,14 +234,14 @@ standard library made a utility for itself called `Unique<T>` which:
...
-| `PhantomData<*const T>`     | -         | covariant                 |
-| `PhantomData<*mut T>`       | -         | invariant                 |
-| `PhantomData<fn(T)>`        | -         | contravariant             |
-| `PhantomData<fn() -> T>`    | -         | covariant                 |
-| `PhantomData<fn(T) -> T>`   | -         | invariant                 |
-| `PhantomData<Cell<&'a ()>>` | invariant | -                         |
+| Phantom type                | `'a`      | `T`                         | `Send`    | `Sync`    |
+|-----------------------------|-----------|-----------------------------|-----------|-----------|
+| `PhantomData<T>`            | -         | covariant (with drop check) | `T: Send` | `T: Sync` |
+| `PhantomData<&'a T>`        | covariant | covariant                   | `T: Send` | `T: Sync` |
```
matthewjasper, line 240: Are these not both `T: Sync`?

Enselic: Yes, good catch, thank you. Fixed.

JohnTitor (approved):

JohnTitor:
Thanks!

PR #413, opened 2023-06-17
Fill "Beneath `std`"

kadiwa4:
[Rendered](https://github.com/kadiwa4/nomicon/blob/beneath_std/src/beneath-std.md)

Based on [this page](https://doc.rust-lang.org/nightly/unstable-book/language-features/lang-items.html) in the Unstable Book. I want to update it (see rust-lang/rust#110274) and I think a lot of information doesn't belong there and is better kept here. The examples can even be tested.

If people don't think that this page fits here, another option would be to create a page for rustc's `start` feature flag (in the Unstable Book) and put the information there.

### DIFF
--- a/src/beneath-std.md
+++ b/src/beneath-std.md
@@ -1,10 +1,108 @@
-# Beneath std
+# Beneath `std`
 
-This section documents (or will document) features that are provided by the standard library and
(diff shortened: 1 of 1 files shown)


kadiwa4:
Both the Rust Book and the error index point to that page now, it would be great if someone found the time to review this.

JohnTitor (approved):
LGTM!

PR #414, opened 2023-07-03
Minor improvements

evaporei:
- Add missing space in lifetime notation (Leaking page)
- Use shorthand notation for field initialization across `new` methods, just like [Arc does currently](https://github.com/rust-lang/nomicon/blob/master/src/arc-mutex/arc-base.md)
- Add missing space in threads execution example (Atomics page)


JohnTitor (approved):
Thanks!

PR #415, opened 2023-07-06
Minor improvements to Vec

evaporei:
- Make both comparisons that call `grow()` use the same order of `len` and `cap`
- Use shorthand notation for field initialization, just as its being used across the book's codebase
- Remove unused import (PhantomData) in the Layout page
- Limit scope of `unsafe` usage to only where it's really needed

I copied the "Final Code" page into a local `cargo new --lib` folder and it compiled and ran tests successfully :heavy_check_mark: (edit: I forgot the CI is running doc tests lol)


JohnTitor (approved):
👍

PR #418, opened 2023-08-31
Fix `Vec::push_all` ptr code in exception-safety

### LINKED ISSUE #359: Potential error in exception-safety
In https://doc.rust-lang.org/nomicon/exception-safety.html I suspect that `self.ptr().add(i).write(x.clone());` is wrong. `i` is the position of `x` in `to_push`. We don't want to put `x` at position `i` as it would overwrite an already existing value. Instead we want it at position `i+initial_length`. 
At least I assume. I don't understand how to find the documentation of `ptr()`, I don't see it in https://doc.rust-lang.org/stable/std/vec/struct.Vec.html . But I can't figure out a semantic off "add" where it makes sens to take "i"

guilliamxavier:
Fix #359

### DIFF
--- a/src/exception-safety.md
+++ b/src/exception-safety.md
@@ -41,11 +41,13 @@ impl<T: Clone> Vec<T> {
     fn push_all(&mut self, to_push: &[T]) {
         self.reserve(to_push.len());
         unsafe {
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):
👍

Issue #135, opened 2019-05-23
Document when we *do* guarnatee that drop runs


RalfJung:
In my understanding, we do in some circumstances guarantee that drop runs. For example:
```rust
struct Guard;

impl Drop for Guard {
    fn drop(&mut self) {
        println!("Hello!");
    }
}

pub fn test(f: impl FnOnce()) {
    let _guard = Guard;
    f();
}
```
Here, we guarantee that no matter the environment or whatever `f` does, *if* the stack frame of `test` ever gets popped or otherwise "deallocated", then the `println!` certainly happens. For example, `f` might loop forever or abort the process, but it cannot "return" or "unwind" or finish in any other way that would circumvent the printing, nor can it use `longjmp` to skip `test`'s cleanup, not can it just ask the OS to outright kill the current thread (without tearing down the entire process).

This is needed to actually realize [the pinning drop guarantee](https://doc.rust-lang.org/nightly/std/pin/index.html#drop-guarantee), but it seems not to be documented anywhere explicitly?

Cc @Gankro @nikomatsakis @comex

Gankro:
I would especially appreciate if someone could properly write down in one of our primary sources (nomicon, trpl, or reference) what on earth happens with double-panics and panic-in-destructors. I feel like I've always received vague hand-waves on this matter.

Gankro:
Note the nomicon discusses some stuff:

* https://doc.rust-lang.org/nightly/nomicon/destructors.html
* https://doc.rust-lang.org/nightly/nomicon/leaking.html

But these docs are getting *extremely* long in the tooth (ManuallyDrop exists now, for instance).

RalfJung:
> I would especially appreciate if someone could properly write down in one of our primary sources (nomicon, trpl, or reference) what on earth happens with double-panics and panic-in-destructors. I feel like I've always received vague hand-waves on this matter.

That would be https://github.com/rust-lang-nursery/reference/issues/348. I see as as somewhat orthogonal to whether unsafe code is allowed to do things like `longjmp` or kill a thread without running the destructors in the stack frames.

Gankro:
The most important thing, imo, is that the "fixed" scoped_thread pattern works:

```rust
fn main() {
  state_to_get_borrowed_by_other_thread = ...;
  scope(|spawner| {
    for i in 0..10 {
      let borrow = &state_to_get_borrowed_by_other_thread;
      spawner.scoped_thread(move || { 
        process(borrow); 
      });
    }
[...]

PR #420, opened 2023-09-06
remove 'fail to call destructors' from okay-list

RalfJung:
This would need a bunch of caveats: when data is pinned, it *is* unsound to not call the destructors (pinning wasn't a thing yet when the nomicon got written); also code can rely on the destructors in its own stack frame to be executed if the stack frame ever gets popped (i.e., longjmp-ing over the stack frame is not okay).


ehuss (approved):
Thanks! I'm going to leave followup on adding all these caveats and gotchas to be tracked in #135.

PR #425, opened 2023-10-07
Handle drop zst

pwbh:
Fix https://github.com/rust-lang/nomicon/issues/424

### DIFF
--- a/src/vec/vec-zsts.md
+++ b/src/vec/vec-zsts.md
@@ -1,13 +1,13 @@
 # Handling Zero-Sized Types
 
 It's time. We're going to fight the specter that is zero-sized types. Safe Rust
-*never* needs to care about this, but Vec is very intensive on raw pointers and
+_never_ needs to care about this, but Vec is very intensive on raw pointers and
 raw allocations, which are exactly the two things that care about
 zero-sized types. We need to be careful of two things:
 
-* The raw allocator API has undefined behavior if you pass in 0 for an
+- The raw allocator API has undefined behavior if you pass in 0 for an
   allocation size.
-* raw pointer offsets are no-ops for zero-sized types, which will break our
+- raw pointer offsets are no-ops for zero-sized types, which will break our
   C-style pointer iterator.
 
 Thankfully we abstracted out pointer-iterators and allocating handling into
@@ -30,6 +30,7 @@ Due to our current architecture, all this means is writing 3 guards, one in each
 method of `RawVec`.
 
 <!-- ignore: simplified code -->
+
 ```rust,ignore
 impl<T> RawVec<T> {
     fn new() -> Self {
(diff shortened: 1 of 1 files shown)


pwbh:
Wonder if possible to merge this :)

pdietl:
@ehuss can this be merged?

pwbh:
@JohnTitor could you take a look please? :)

src/vec/vec-zsts.md
```diff
@@ -214,3 +214,20 @@ impl<T> DoubleEndedIterator for RawValIter<T> {
 ```
 
 And that's it. Iteration works!
+
+One last thing that we need to take into account is that now when our vec gets dropped it deallocates the memory that was allocated during the time our vec was alive. With ZSTs we did not allocate any memory, and infact we never do. So right now we have unsoundness in our code where we still try deallocate a `NonNull::dangling()` ptr that we use to simulate the ZST in our vec, This means that we would cause an undefined behaviour if we try to deallocate something that we never allocated (obviously and for the right reasons). Lets fix tha, in our raw_vec we are going to tweak our Drop trait and check that we deallocate only types that are sized.
```
baumanj, line 218:
```suggestion
One last thing that we need to take into account is that now when our vec gets dropped it deallocates the memory that was allocated during the time our vec was alive. With ZSTs we did not allocate any memory, and in fact we never do. So right now we have unsoundness in our code where we still try deallocate a `NonNull::dangling()` ptr that we use to simulate the ZST in our vec, This means that we would cause undefined behavior if we try to deallocate something that we never allocated (obviously and for the right reasons). To fix that, in our `RawVec` we are going to tweak our `Drop` trait and check that we deallocate only types that are sized.
```

By the way, there's nothing wrong with the "behaviour" spelling, but this repo consistently uses "behavior".

pwbh:
@baumanj merged your suggestion, thanks!

src/vec/vec-zsts.md
```diff
@@ -214,3 +214,20 @@ impl<T> DoubleEndedIterator for RawValIter<T> {
 ```
 
 And that's it. Iteration works!
+
+One last thing that we need to take into account is that now when our vec gets dropped it deallocates the memory that was allocated during the time our vec was alive. With ZSTs we did not allocate any memory, and in fact we never do. So right now we have unsoundness in our code where we still try deallocate a `NonNull::dangling()` ptr that we use to simulate the ZST in our vec, This means that we would cause undefined behavior if we try to deallocate something that we never allocated (obviously and for the right reasons). To fix that, in our `RawVec` we are going to tweak our `Drop` trait and check that we deallocate only types that are sized.
```
JohnTitor, line 218:
```suggestion
One last thing we need to consider is that when our vector is dropped, it deallocates the memory that was allocated while it was alive. With ZSTs, we didn't allocate any memory; in fact, we never do. So, right now, our code has unsoundness: we're still trying to deallocate a `NonNull::dangling()` pointer that we use to simulate the ZST in our vector. This means we'd cause undefined behavior if we tried to deallocate something we never allocated (obviously, and for good reasons). To fix this, in our `RawVec`'s `Drop` trait, we're going to tweak it to ensure we only deallocate types that are sized.
```

PR #427, opened 2023-10-17
Fixed `Hole::get` marked as unsafe in `exception-safety.md`

### LINKED ISSUE #426: `Exception Safety` section: `Hole::get` probably doesn't have to be `unsafe`
https://doc.rust-lang.org/nomicon/exception-safety.html

```rust
struct Hole<'a, T: 'a> {
    data: &'a mut [T],
    /// `elt` is always `Some` from new until drop.
    elt: Option<T>,
    pos: usize,
}

impl<'a, T> Hole<'a, T> {
    // ...

    unsafe fn get(&self, index: usize) -> &T { &self.data[index] }

    // ...
}
```

Why is `get` here marked as `unsafe`? It doesn't seem to be performing anything particularly unsafe, any more so than regular safe code might do. Rust compiler also doesn't require it to be `unsafe`.

frozenspider:
Resolves #426

### DIFF
--- a/src/exception-safety.md
+++ b/src/exception-safety.md
@@ -172,7 +172,7 @@ impl<'a, T> Hole<'a, T> {
 
     fn removed(&self) -> &T { self.elt.as_ref().unwrap() }
 
-    unsafe fn get(&self, index: usize) -> &T { &self.data[index] }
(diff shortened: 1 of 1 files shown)


JohnTitor:
Thank you!

PR #429, opened 2023-10-28
typo: probably

krystian-wojtas:


ehuss:
Thanks for the PR!  I believe the original is correct. It is saying that the compiler will automatically implement those traits when it can prove that it is safe using the rules of the language to generate the proof.

JohnTitor:
Yeah, it should be prove+ably, closing as wontfix. Thanks for the PR anyway!

PR #431, opened 2023-11-19
Reword the section on general race conditions

tgross35:
The section on preventing general race conditions is a bit hand wavy. Change wording to be more concrete, and add an example of Rust preventing general races in a very specific case.


ehuss (approved):
Thanks!

PR #439, opened 2024-01-16
add a funny

TheBearodactyl:
pretty sure this'll get closed, but i might as well try lmfao


ehuss:
Thanks, I appreciate the humor 😄, but yea I don't think we want to link to an external site.

TheBearodactyl:
> Thanks, I appreciate the humor 😄, but yea I don't think we want to link to an external site.

yeah, i kinda guessed. worth a shot though :D

PR #440, opened 2024-01-16
Add build script part to FFI chapter for more clear and smooth learn …

### LINKED ISSUE #122: Provide working example for FFI
It took me a long time to get the "snappy" example, included in the FFI section, working on my machine. As it turned out, I needed several options in a `build.rs` file--one of which I didn't find until I stumbled across the [rust-snappy](https://github.com/paritytech/rust-snappy) repository.

Namely, I needed the following `build.rs` file:

```rust
fn main() {
    println!("cargo:rustc-link-search=native=/path/to/snappy/");
    println!("cargo:rustc-link-lib=static=snappy");
    println!("cargo:include=/path/to/snappy/");

    // Need to link C with C++ files
    println!("cargo:rustc-link-lib=stdc++");
}
```

This is never mentioned in the FFI section, and it was a bit frustrating. I've seen it mentioned several places that it's "easy to call C from Rust", and I think giving people everything they need to know to be successful in the Rustonomicon would help make that even more true.

In addition, specifically mentioning the necessity of the `stcd++` library for linking to C wrappers around C++ libraries would be great. I'm sure a lot of people are already familiar with that concept, but many others (myself included) don't know about it and would need a lot of trial and error to get that to work.

Do others agree? I'm willing to attempt a write-up if I could get some feedback on where it would be appropriate to include. My thought is to put it near the beginning so that people can get a working example from the beginning, but I'm probably biased since I just went through the experience of being confused.

### LINKED ISSUE #345: Snappy ffi example uses u8 instead of c_char
The ffi interface in the snappy examples doesn't match the c-library types in snappy-c.h.

Should the ffi interface use ```libc::c_char``` instead of ```u8```

https://github.com/google/snappy/blob/main/snappy-c.h#L71
```
   snappy_status snappy_compress(const char* input,
                              size_t input_length,
                              char* compressed,
                              size_t* compressed_length);
```                              

example here
https://doc.rust-lang.org/nomicon/ffi.html#calling-foreign-functions
```
   fn snappy_compress(input: *const u8,
                       input_length: size_t,
                       compressed: *mut u8,
                       compressed_length: *mut size_t) -> c_int;
```                       

Would it be clearer to have these function definition match then cast the pointer inside the rust function.

RoggeOhta:
Fix #345, fix #122 
Make FFI chapter snappy example workable.

### DIFF
--- a/src/ffi.md
+++ b/src/ffi.md
@@ -36,8 +36,8 @@ Then create a new file at the root of your workspace, named `build.rs`.
 ```rust
 // build.rs
 fn main() {
-    println!("cargo:rustc-link-lib=dylib=stdc++");
-    println!("cargo:rustc-link-search=<YOUR SNAPPY LIBRARY PATH>");
+    println!("cargo:rustc-link-lib=dylib=stdc++");
+    println!("cargo:rustc-link-search=<YOUR SNAPPY LIBRARY PATH>");
 }
 ```
 
--- a/src/ffi.md
+++ b/src/ffi.md
@@ -21,6 +21,29 @@ libc = "0.2.0"
 
 [libc]: https://crates.io/crates/libc
 
+## Prepare the build script
+
+Because [snappy](https://github.com/google/snappy) is a static library by default. So there is no C++ std linked in the output artifact. In order to use this foreign library in Rust, we have to manually specify that we want to link stdc++ in our project. The easiest way to do this is by setting up a build script.
+
+First edit `Cargo.toml`, inside `package` add `build = "build.rs"`.
+```toml
+[package]
+...
+build = "build.rs"
+```
+
+Then create a new file at the root of your workspace, named `build.rs`.
+```rust
+// build.rs
+fn main() {
+    println!("cargo:rustc-link-lib=dylib=stdc++");
+    println!("cargo:rustc-link-search=<YOUR SNAPPY LIBRARY PATH>");
+}
+```
+
+For more information, please read [The Cargo Book - build script](https://doc.rust-lang.org/cargo/reference/build-scripts.html).
+
+
 ## Calling foreign functions
 
 The following is a minimal example of calling a foreign function which will
--- a/src/ffi.md
+++ b/src/ffi.md
@@ -23,20 +23,23 @@ libc = "0.2.0"
 
 ## Prepare the build script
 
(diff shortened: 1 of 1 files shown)


jamischarles:
@RoggeOhta  Thank you thank you thank you. I've been banging my head wondering how on earth to reference snappy.

jamischarles:
FYI: on macOS I didn't need `println!("cargo:rustc-link-lib=dylib=stdc++");` this line in the build script.
In fact it caused an error
`note: ld: library 'stdc++' not found`

jamischarles:
My working build file:
```rust
fn main() {
    // println!("cargo:rustc-link-lib=dylib=stdc++");
    println!("cargo:rustc-link-search=/Users/jacharles/dev_freelance/snappy/build");
}
```

src/ffi.md
```diff
@@ -21,6 +21,29 @@ libc = "0.2.0"
 
 [libc]: https://crates.io/crates/libc
 
+## Prepare the build script
+
+Because [snappy](https://github.com/google/snappy) is a static library by default. So there is no C++ std linked in the output artifact. In order to use this foreign library in Rust, we have to manually specify that we want to link stdc++ in our project. The easiest way to do this is by setting up a build script.
```
JohnTitor, line 26:
```suggestion
Because [snappy](https://github.com/google/snappy) is a static library by default.
So there is no C++ std linked in the output artifact. 
n order to use this foreign library in Rust, we have to manually specify that we want to link stdc++ in our project.
The easiest way to do this is by setting up a build script.
```

src/ffi.md
```diff
@@ -21,6 +21,29 @@ libc = "0.2.0"
...
+
+First edit `Cargo.toml`, inside `package` add `build = "build.rs"`.
```
JohnTitor, line 28:
```suggestion
First edit `Cargo.toml`, inside `package` add `build = "build.rs"`:
```

src/ffi.md
```diff
@@ -21,6 +21,29 @@ libc = "0.2.0"
...
+```toml
+[package]
+...
+build = "build.rs"
+```
+
+Then create a new file at the root of your workspace, named `build.rs`.
```
JohnTitor, line 35:
```suggestion
Then create a new file at the root of your workspace, named `build.rs`:
```

src/ffi.md
```diff
@@ -21,6 +21,29 @@ libc = "0.2.0"
...
+```rust
+// build.rs
+fn main() {
+    println!("cargo:rustc-link-lib=dylib=stdc++");
```
JohnTitor, line 39:
```suggestion
    println!("cargo:rustc-link-lib=dylib=stdc++"); // This line may be unnecessary for some environment.
```

PR #443, opened 2024-03-16
Cleanups for tree example of splitting borrows

ratmice:
This just does some cleanups of the code in the split borrows tree example, some suggested by clippy
others just seemed like simplifications that could be done.


JohnTitor (approved):
Thanks!

PR #446, opened 2024-03-21
Fixed typo 'analyses' -> 'analysis'

kaffarell:
changed 'analyses' to 'analysis'.


ehuss:
Can you say more about why this seems to be a typo? It looks correct to me, as both analyses and optimizations are plural.

kaffarell:
Oops, didn't notice it was in plural, though it was meant to be singular :) never mind then.

PR #450, opened 2024-05-15
Tidy up beneath-std, support Windows-MSVC without libc

saethlin:
I've run into this because my other PR breaks these examples: https://github.com/rust-lang/rust/pull/124050

(also the comment about `crt0.o` on these can't be right, because this was being built on Windows)

### DIFF
--- a/src/beneath-std.md
+++ b/src/beneath-std.md
@@ -19,7 +19,10 @@ Note that the default features have been disabled. This is a critical step -
 disabled.**
 
 Alternatively, we can use the unstable `rustc_private` private feature together
-with an `extern crate libc;` declaration as shown in the examples below.
+with an `extern crate libc;` declaration as shown in the examples below. Note that
+windows-msvc targets do not require a libc, and correspondingly there is no `libc`
+crate in their sysroot. We do not need the `extern crate libc;` below, and having it
+on a windows-msvc target would be a compile error.
 
 ## Writing an executable without `std`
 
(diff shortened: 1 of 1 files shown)


bjorn3:
On Unix a dependency on libc is still necessary for crt1.o. As is libunwind for any panic=unwind target (even when compiling with panic=abort, as libcore references it _Unwind_Resume from libunwind)

saethlin:
Ah! I'll improve the comments.

src/beneath-std.md
```diff
@@ -39,11 +42,12 @@ in the same format as C (aside from the exact integer types being used):
 #![allow(internal_features)]
 #![no_std]
 
-// Necessary for `panic = "unwind"` builds on some platforms.
+// Necessary for `panic = "unwind"` builds on cfg(unix) platforms.
 #![feature(panic_unwind)]
 extern crate unwind;
 
 // Pull in the system libc library for what crt0.o likely requires.
+#[cfg(not(windows))]
```
bjorn3, line 50: I suspect windows-gnu still needs it, but I'm not entirely sure.

ChrisDenton: windows-gnu includes gnu libc [via the compiler](https://github.com/rust-lang/rust/blob/6579ed89f0fcc26da71afdd11d30d63f6f812a0a/compiler/rustc_target/src/spec/base/windows_gnu.rs#L30). The libc crate doesn't add anything, as far as I'm aware.

src/beneath-std.md
```diff
@@ -19,7 +19,10 @@ Note that the default features have been disabled. This is a critical step -
 disabled.**
 
 Alternatively, we can use the unstable `rustc_private` private feature together
-with an `extern crate libc;` declaration as shown in the examples below.
+with an `extern crate libc;` declaration as shown in the examples below. Note that
+windows-msvc targets do not require a libc, and correspondingly there is no `libc`
+crate in their sysroot. We not need the `extern crate libc;` below, and having it
```
RalfJung, line 24:
```suggestion
crate in their sysroot. We do not need the `extern crate libc;` below, and having it
```
or something like that?

saethlin: Oops, yeah your fix is correct.

ehuss (approved):
Thanks!

PR #455, opened 2024-06-19
expand leak amplification with example

kristof-mattei:
Took me quite a while to understand leak amplification actually meant until I tested it in code.

I think this makes it more clear.


LeoniePhiline:
Maybe your example preface could mention where the actual _amplification_ happens? 

Only plain leaking is mentioned.

PR #458, opened 2024-08-03
repr(int) enums: both size and sign matter

RalfJung:
The text makes it sound like only the size matters, but `i32` and `u32` are not ABI-compatible so we have to talk about size and sign here.

### DIFF
--- a/src/other-reprs.md
+++ b/src/other-reprs.md
@@ -42,7 +42,7 @@ says they should still consume a byte of space.
   difference from a struct is that the fields aren’t named.
 
 * `repr(C)` is equivalent to one of `repr(u*)` (see the next section) for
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

Might it also be a good idea to update https://github.com/rust-lang/reference/blob/72267820df5533a75505a81a97dba670334f0d57/src/type-layout.md#reprc-field-less-enums with the same wording?

RalfJung:
> Might it also be a good idea to update https://github.com/rust-lang/reference/blob/72267820df5533a75505a81a97dba670334f0d57/src/type-layout.md#reprc-field-less-enums with the same wording?

Not sure -- that file is largely about the type's *layout*, which would not include the sign. The sign only matters for function call ABI. A file titled "type layout" seems like the wrong place to discuss that.

PR #460, opened 2024-08-15
other-reprs: Add details for n!=1 repr(packed)

luqmana:
Expand the `repr(packed)` section to note if also supports an explicit power-of-two argument like `repr(align(n))`.


JohnTitor (approved):
Looks good, thanks!

PR #462, opened 2024-09-22
Intro: fix word order

ttrssreal:
Is it this or is it "how constructs lower to compiler/OS/hardware work".


ehuss:
I believe the original is correct. "Constructs" in this case can be interpreted as a plural noun which essentially means "things in the language", and "lower" is the verb, which essentially means *translated*. "Lowering" is a term used in compilers when translating from a higher-level representation to a lower-level one.

ttrssreal:
Oh I see, sorry :3

Issue #198, opened 2020-02-07
Apparent contradiction between nomicon and reference docs for ZST pointers


Lucretiel:
There's a [passage](https://doc.rust-lang.org/nomicon/vec-zsts.html#allocating-zero-sized-types) of the nomicon that reads:

> So if the allocator API doesn't support zero-sized allocations, what on earth do we store as our allocation? Unique::empty() of course! Almost every operation with a ZST is a no-op since ZSTs have exactly one value, and therefore no state needs to be considered to store or load them. This actually extends to ptr::read and ptr::write: they won't actually look at the pointer at all. As such we never need to change the pointer.

However, the docs for [`ptr::read`](https://doc.rust-lang.org/std/ptr/fn.read.html#safety) state:

> Note that even if T has size 0, the pointer must be non-NULL and properly aligned.

Is this a contradiction? The former passage states that `ptr::read` for a ZST is a "no-op" that "won't actually look at the pointer at all", which would seem to imply it doesn't matter what data you pass to it, but the latter passage states that it DOES matter, you must pass an aligned, non-null pointer.

SNCPlay42:
As far as I know it's likely correct that `ptr::read`/`write` on ZSTs do nothing *when* they're well defined, so the nomicon isn't entirely incorrect to call them "no-ops". But that doesn't mean they *aren't* undefined for null or unaligned pointers. So there sort of is and isn't a contradiction, depending on how you read it - by "they won't actually look at the pointer", does the nomicon mean "they won't look at the *pointed memory*", or "they won't look at the *actual pointer*"? It should definitely avoid implying the latter, so the wording should be changed.

It's worth noting that `Unique::empty()` (and its stable equivalent [`NonNull::dangling()`](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.dangling)) is a non-null, properly aligned pointer, so the actual code in the nomicon shouldn't be broken.

nico-abram:
Should the it also reference NonNull::dangling instead of Unique::empty?

PR #467, opened 2024-10-15
Say that dereferencing a pointer to a ZST is no longer undefined

mattheww:
The new rules were tracked in
https://github.com/rust-lang/rust/issues/117945

The corresponding update to the Reference was
https://github.com/rust-lang/reference/pull/1541

See also #198

### DIFF
--- a/src/exotic-sizes.md
+++ b/src/exotic-sizes.md
@@ -106,9 +106,9 @@ consequence of types with no size. In particular, pointer offsets are no-ops,
 and allocators typically [require a non-zero size][alloc].
 
 Note that references to ZSTs (including empty slices), just like all other
-references, must be non-null and suitably aligned. Dereferencing a null or
-unaligned pointer to a ZST is [undefined behavior][ub], just like for any other
-type.
+references, must be non-null and suitably aligned. However, dereferencing a
+null pointer to a ZST is not [undefined behavior][ub], unlike pointers to
+other types.
 
 [alloc]: ../std/alloc/trait.GlobalAlloc.html#tymethod.alloc
 [ub]: what-unsafe-does.html
--- a/src/exotic-sizes.md
+++ b/src/exotic-sizes.md
@@ -106,9 +106,9 @@ consequence of types with no size. In particular, pointer offsets are no-ops,
 and allocators typically [require a non-zero size][alloc].
 
 Note that references to ZSTs (including empty slices), just like all other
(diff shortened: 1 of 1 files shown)


src/exotic-sizes.md
```diff
@@ -106,9 +106,9 @@ consequence of types with no size. In particular, pointer offsets are no-ops,
 and allocators typically [require a non-zero size][alloc].
 
 Note that references to ZSTs (including empty slices), just like all other
-references, must be non-null and suitably aligned. Dereferencing a null or
-unaligned pointer to a ZST is [undefined behavior][ub], just like for any other
-type.
+references, must be non-null and suitably aligned. However, dereferencing a
+null pointer to a ZST is not [undefined behavior][ub], unlike pointers to
+other types.
```
RalfJung, line 111: In fact dereferencing (as in, using the `*` operator on) a null pointer is never UB. `addr_of!(*ptr)` is always safe, even if `ptr` is null.

Only loads, stores, and field projections have any requirements.

mattheww: I've changed it to say "loading or storing through a null pointer to a ZST"

src/exotic-sizes.md
```diff
@@ -106,9 +106,9 @@ consequence of types with no size. In particular, pointer offsets are no-ops,
 and allocators typically [require a non-zero size][alloc].
 
 Note that references to ZSTs (including empty slices), just like all other
-references, must be non-null and suitably aligned. Dereferencing a null or
-unaligned pointer to a ZST is [undefined behavior][ub], just like for any other
-type.
+references, must be non-null and suitably aligned. However, loading or storing
+through a null pointer to a ZST is not [undefined behavior][ub], unlike
+pointers to other types.
```
RalfJung, line 111: I think it would be less confusing to say something like "loading or storing through a null pointer is UB, unless the load/store has size 0". Let's talk about the common case first.

mattheww: The subject of this section is ZSTs, so "loading or storing through a null pointer to a ZST is not undefined behavior" is the information it's supplying, and "unlike pointers to other types" is an aside.

RalfJung: Ah, fair.

traviscross (approved):

traviscross:
The corresponding lang FCP was:

- https://github.com/rust-lang/rust/pull/117329

PR #468, opened 2024-10-28
races: Clarify a “mostly” that might be misread

andersk:
One might carelessly misread a sentence that begins “data races are *mostly* prevented” as suggesting that Rust fails to prevent some data races.  Clarify the intended reading.


JohnTitor (approved):
LGTM, thanks!

PR #470, opened 2024-12-05
races.md: data race -> race condition to violate memory safety

### LINKED ISSUE #466: Which one does the last code in 8.1. Races cause, a data race or a race condition?
The comment in the last code in [8.1. Races](https://doc.rust-lang.org/nomicon/races.html) says "This is a race condition", but the description right above says "We can cause a data race". Which one is correct?

I guess it doesn't cause a data race, because all of the accesses to the `AtomicUsize` are synchronized, so the third requirement of a data race "one or more of them is unsynchronized" is not met.

Enselic:
The first example shows that you "can't violate memory safety with safe Rust" and the second example shows that you "can violate memory safety with unsafe Rust".

The second example does not demonstrate a data race since there is only one thread touching `data`.

Fixes #466

### DIFF
--- a/src/races.md
+++ b/src/races.md
@@ -60,8 +60,8 @@ thread::spawn(move || {
 println!("{}", data[idx.load(Ordering::SeqCst)]);
 ```
 
-We can cause a data race if we instead do the bound check in advance, and then
-unsafely access the data with an unchecked value:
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):
Thanks!

PR #471, opened 2024-12-15
Fix typo

wtlin1228:


ehuss:
Thanks for the PR! The original looks correct to me. It is referring to *uses* (plural) and there can be multiple uses, of which there can be moves or drops (plural).

wtlin1228:
Got it. Thanks for the review.

PR #476, opened 2025-01-19
Update guidance on uninitialized fields to use &raw mut instead of addr_of_mut!

nikthechampiongr:
The doc page on [addr_of_mut](https://doc.rust-lang.org/std/ptr/macro.addr_of_mut.html) and [addr_of](https://doc.rust-lang.org/std/ptr/macro.addr_of.html) state that the macros are soft deprecated in favor of &raw mut/&raw const.


ehuss (approved):
Thanks!

PR #478, opened 2025-01-23
Remove `#![start]` attribute

ehuss:
The `#![start]` attribute has been removed in
https://github.com/rust-lang/rust/pull/134299.

### DIFF
--- a/src/beneath-std.md
+++ b/src/beneath-std.md
@@ -30,46 +30,9 @@ We will probably need a nightly version of the compiler to produce
 a `#![no_std]` executable because on many platforms, we have to provide the
 `eh_personality` [lang item], which is unstable.
 
-Controlling the entry point is possible in two ways: the `#[start]` attribute,
-or overriding the default shim for the C `main` function with your own.
+You will need to define a symbol for the entry point that is suitable for your target. For example, `main`, `_start`, `WinMain`, or whatever starting point is relevant for your target. Additionally, we use the `#![no_main]` attribute to override the compiler-inserted `main` shim.
+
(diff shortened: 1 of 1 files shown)


ehuss:
@RalfJung Can you please give this a review?

src/beneath-std.md
```diff
@@ -30,46 +30,9 @@ We will probably need a nightly version of the compiler to produce
...
-Additionally, it's required to define a [panic handler function](panic-handler.html).
-
-The function marked `#[start]` is passed the command line parameters
-in the same format as C (aside from the exact integer types being used):
-
-```rust
-#![feature(start, lang_items, core_intrinsics, rustc_private)]
-#![allow(internal_features)]
-#![no_std]
+You will need to define a symbol for the entry point that is suitable for your target. For example, `main`, `_start`, `WinMain`, or whatever starting point is relevant for your target. Additionally, we use the `#![no_main]` attribute to override the compiler-inserted `main` shim.
```
RalfJung, line 33:
```suggestion
You will need to define a symbol for the entry point that is suitable for your target. For example, `main`, `_start`, `WinMain`, or whatever starting point is relevant for your target.
Additionally, you need to use the `#![no_main]` attribute to prevent the compiler from attempting to generate an entry point itself.
```

RalfJung (commented):
I don't know which starting points must be used where, `main` worked for everything except wasm in my PR. I guess `WinMain` is specific for GUI applications on Windows, no idea where `_start` would be used.

traviscross (approved):

ehuss:
Thanks @RalfJung!

PR #481, opened 2025-03-01
Update to 2024 edition

printfn:
I noticed this was still using Rust 2021, so I went and updated it for the 2024 edition.


JohnTitor (approved):
Thanks!

PR #485, opened 2025-04-07
remove `rust-intrinsic` ABI

Skgland:
The rust-intrinsic ABI has been removed from the compiler in rust-lang/rust#139455 as part of rust-lang/rust#132735


ehuss (approved):
Thanks!

PR #486, opened 2025-04-08
clarify passage on problematic use of transmute

createyourpersonalaccount:


traviscross:
Thanks for the PR.

I kind of see what this is trying to get at, but it would need to say a lot more to make this point correctly.  And it may not be worth doing that since the elaboration is already there on the linked pages.  So we're going to close this.

createyourpersonalaccount:
@traviscross The problem is the verb **produces**, as it's not always true. Why not be correct in technical documentation? Don't let perfect be the enemy of good.

traviscross:
Maybe it'd help if you wrote out a code example of the case you're trying to cover here.

createyourpersonalaccount:
@traviscross I have a `&mut [MaybeUninit<u8>]` that I want to use as a read buffer:

```rust
unsafe { transmute::<&mut [MaybeUninit<u8>], &mut [u8]>(&mut buf) }
```

Now this value is passed into `.read()`. As I understand it, there's no explicit lifetime here, it's part of a struct method, so the elision rules apply and it is not an unbounded lifetime. When I read the old passage (that I'm trying to correct) I had to think twice and ask around to make sure that I'm not making a mistake, because indeed "produces" means in all cases, and as far as I could tell, the textualist interpretation says my code is wrong.

traviscross:
If you could write out a full example including the call to `read` and the enclosing function, that'd be helpful, as then I can explain in terms of that.

createyourpersonalaccount:
@traviscross Here is an example, https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=93d67518aa80dac37a8cfca1d612140b unfortunately I can't make it compile there because I'm using the [`bytes_deque`](https://crates.io/crates/bytes_deque) crate, which is not in the top 100 so playground doesn't support it. You'll have to test locally.

traviscross:
In your example, the lifetime of that slice is indeed unbounded.  To see this, add, e.g.:

```rust
        fn is_static<T: 'static>(_: T) {}
        is_static(slice);
```

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&code=use+bytes%3A%3ABytesMut%3B%0Ause+std%3A%3Aio%3A%3A%7Bself%2C+Read%7D%3B%0Ause+std%3A%3Amem%3A%3Atransmute%3B%0A%0Apub+struct+S%3CR%3A+Read%3E+%7B%0A++++pub+r%3A+R%2C%0A++++pub+b%3A+BytesMut%2C%0A%7D%0A%0Aimpl%3CR%3A+Read%3E+S%3CR%3E+%7B%0A++++pub+fn+read_more%28%26mut+self%29+-%3E+io%3A%3AResult%3Cusize%3E+%7B%0A++++++++let+slice%3A+%26mut+%5Bu8%5D+%3D%0A++++++++++++unsafe+%7B+transmute%28self.b.spare_capacity_mut%28%29%29+%7D%3B%0A++++++++let+result+%3D+self.r.read%28slice%29%3B%0A++++++++if+let+Ok%28m%29+%3D+result+%7B%0A++++++++++++unsafe+%7B%0A++++++++++++++++self.b.set_len%28self.b.len%28%29+%2B+m%29%3B%0A++++++++++++%7D%0A++++++++%7D%0A++++++++fn+is_static%3CT%3A+%27static%3E%28_%3A+T%29+%7B%7D%0A++++++++is_static%28slice%29%3B%0A++++++++return+Ok%280%29%3B%0A++++%7D%0A%7D%0A%0Afn+main%28%29+%7B%0A++++let+mut+st+%3D+S+%7B%0A++++++++r%3A+b%22world%22.as_slice%28%29%2C%0A++++++++b%3A+BytesMut%3A%3Afrom%28b%22hello+goodbye%21%22.as_slice%28%29%29%2C%0A++++%7D%3B%0A++++unsafe+%7B+st.b.set_len%286%29+%7D%3B%0A++++st.read_more%28%29.unwrap%28%29%3B%0A++++println%21%28%22%7B%3A%3F%7D%22%2C+st.b%29%3B%0A%7D%0A)

That doesn't necessarily mean you're doing anything wrong, but the lifetime there is not bounded.

createyourpersonalaccount:
I see what you mean. I tried to think of more involved examples to make my point, but I realized that all my valid examples would end up coercing the lifetime, either implicitly or explicitly, from unbounded to named. So I suppose the wording is correct: transmute will produce unbounded lifetimes if that's the types given to it.

Are you still sure that no clarification is needed? Perhaps one more sentence can be added. Every other bullet point has at least two sentences. "The transmuted value should not be used in contexts where the original lifetime would be rejected." or so?

PR #487, opened 2025-04-09
Remove double wording in opaque type chapter

junlarsen:
Fixes up some word soup I came across reading the chapter.


ehuss (approved):
Thanks!

PR #491, opened 2025-05-22
Fix code sample output in unchecked-uninit.md

senekor:
`dbg!()` prints to stderr, which I assume is the reason there is no visible output when running the example in the book. By using `println!()` instead, the output becomes visible.


JohnTitor (approved):
👍

PR #523, opened 2026-06-03
Clarify when safety may rely on correctness

ia0:
This is a follow-up of #294 which in my opinion did not completely address the confusion.

The "Safe Rust client" terminology comes from the [Reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#r-undefined.soundness). I'm using "dependencies" (think upstream code) as the opposite concept of "clients" (think downstream code). I'm happy to follow any suggestions to improve those terms.

### DIFF
--- a/src/safe-unsafe-meaning.md
+++ b/src/safe-unsafe-meaning.md
@@ -58,12 +58,13 @@ interfaces built on top of these implementations can be assumed to be safe.
 The need for all of this separation boils down to a single fundamental property
 of Safe Rust, the *soundness property*:
 
-**No matter what, Safe Rust can't cause Undefined Behavior.**
+**No matter what, Safe Rust clients can't cause Undefined Behavior.**
 
 The design of the safe/unsafe split means that there is an asymmetric trust
 relationship between Safe and Unsafe Rust. Safe Rust inherently has to
 trust that any Unsafe Rust it touches has been written correctly.
-On the other hand, Unsafe Rust cannot trust Safe Rust without care.
+On the other hand, Unsafe Rust cannot trust Safe Rust without care. It can
+trust Safe Rust dependencies but cannot trust Safe Rust clients.
 
 As an example, Rust has the [`PartialOrd`] and [`Ord`] traits to differentiate
 between types which can "just" be compared, and those that provide a "total"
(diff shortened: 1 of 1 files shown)


src/safe-unsafe-meaning.md
```diff
@@ -58,12 +58,13 @@ interfaces built on top of these implementations can be assumed to be safe.
...
 
-**No matter what, Safe Rust can't cause Undefined Behavior.**
+**No matter what, Safe Rust clients can't cause Undefined Behavior.**
 
 The design of the safe/unsafe split means that there is an asymmetric trust
 relationship between Safe and Unsafe Rust. Safe Rust inherently has to
 trust that any Unsafe Rust it touches has been written correctly.
-On the other hand, Unsafe Rust cannot trust Safe Rust without care.
+On the other hand, Unsafe Rust cannot trust Safe Rust without care. It can
+trust Safe Rust dependencies but cannot trust Safe Rust clients.
```
JohnTitor: Maybe it'd be better to describe what cannot be trusted a bit more specific.

```suggestion
On the other hand, Unsafe Rust cannot trust Safe Rust without care. It can
trust specific Safe Rust implementations, but not arbitrary Safe Rust chosen
or supplied by clients.
```

src/safe-unsafe-meaning.md
```diff
@@ -91,8 +92,13 @@ can be weighed against the benefit. In this case there's basically zero risk;
 if integers and slices are broken, *everyone* is broken. Also, they're maintained
 by the same people who maintain `BTreeMap`, so it's easy to keep tabs on them.
 
-On the other hand, `BTreeMap`'s key type is generic. Trusting its `Ord` implementation
-means trusting every `Ord` implementation in the past, present, and future.
+This difference also holds for arbitrary implementations of one very specific
+dependency. Unsafe Rust in crate `foo` (which depends on crate `bar`) may rely on
+Safe Rust in crate `bar` to be written correctly, regardless of the actual
+implementation.
```
JohnTitor: How about this wording? I'd avoid using two words, dependency and crate in the same paragraph.

```suggestion
The same can be true across crate boundaries. Unsafe Rust in crate `foo` may
trust a specific Safe Rust implementation in crate `bar` when that is a
bounded piece of code that can be reviewed.
```

src/safe-unsafe-meaning.md
ia0: What do you mean by "implementation"? My understanding is that it means a specific version of the code that implements an API. If that's true then Unsafe Rust can trust arbitrary implementations of the crates it depends on, it doesn't need to be a specific implementation.

If we want to be more specific, this could look like:
- Unsafe Rust can trust Safe Rust within its own crate (because the implementation is known and thus can be verified to be correct) and the crates of its dependencies (because, while unknown, the implementation is trusted to be correct).
- Unsafe Rust cannot trust Safe Rust within the crates of its reverse-dependencies (because the implementation is unknown and not trusted to be correct).

Not sure how to formulate this elegantly.

src/safe-unsafe-meaning.md
ia0: The wording is much better, but the "specific implementation" issue from the other comment is still present. So I'll wait until we clarify that to update this part. Maybe "implementation" is more a general concept than I think.

ia0 (commented):
Thanks for the review! Before applying the suggestions, I have one clarification question in one of the comment.

src/safe-unsafe-meaning.md
ia0: I reformulated by avoiding both the word "dependencies" and "implementations", simply using the word "clients" in the opposite direction. Does this sound better?

JohnTitor: Yeah the current looks great! What I meant is a bounded body of code that the unsafe code has chosen to rely on as part of its trusted base.

JohnTitor (approved):
Thank you and sorry for the delay!

ia0:
Thanks for the review!

